PR #17, opened 2023-05-24
Use HTTPS URL for the submodule

leandro-lucarella-frequenz:
- Fix RELEASE_NOTES template
- Clear release notes
- Use HTTPS to retrieve the submodule


tiyash-basu-frequenz (approved):

leandro-lucarella-frequenz:
For more context, I bumped into this while testing locally and not having my ssh agent running, I was asked for a password to unlock my key (even when it is not necessary).  I guess the CI works fine because if there is no private key there is nothing to unlock, but there is also no reason not to use https :)

PR #41, opened 2023-08-01
Configure the protobuf include paths

llucax:
By default repo-config adds this repository (api-common) as a dependency when generating new API repos, as it is supposed to be included by most (if not all) APIs. So the default `include_paths` repo-config uses looks like `["submodules/api-common-protos", "submodules/frequenz-api-common/proto"]`.

Of course in this repository we don't have ourself as a cyclic dependency, and we only depend on `submodules/api-common-protos`, so we need to override the `include_paths` option, otherwise we get errors about `submodules/frequenz-api-common/proto` not existing.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -82,6 +82,9 @@ profile = "black"
 line_length = 88
 src_paths = ["benchmarks", "examples", "src", "tests"]
 
+[tool.frequenz-repo-config.protobuf]
+include_paths = ["submodules/api-common-protos"]
+
 [tool.pylint.similarities]
 ignore-comments = ['yes']
 ignore-docstrings = ['yes']


llucax:
To really work this also needs https://github.com/frequenz-floss/frequenz-repo-config-python/pull/101, but better to have it here sooner than later.

Marenz:
> By default repo-config will also include this repository in the include paths, as other API repos will depend on it, so we need to make an exception here and only include the google stuff.

Which repository is "this repository"?
What kind of exception are we adding? What is the effect of it?

llucax:
This repo is this repo 😂 (api-common)

llucax:
OK, point taken, the text is not clear enough for people not familiar with the details about protobuf handling in repo-config, I'll try to clarify.

llucax:
Updated commit and PR messages.

Marenz (approved):

llucax:
This still needs the approval from someone on the @frequenz-floss/api-team

tiyash-basu-frequenz (approved):

PR #90, opened 2023-09-26
Add versioning to common proto files

### LINKED ISSUE #42: Introduce package based versioning
### What's needed?

The common API definitions should make use of package based versioning.

### Proposed solution

See https://cloud.google.com/apis/design/versioning

### Use cases

_No response_

_No response_

### Additional context

_No response_

tiyash-basu-frequenz:
This commit adds versioning to the common proto files. This is done by adding a v1 directory to the common directory and moving all the proto files into it. The proto packages are also renamed to include the v1 suffix.

This commit also updates the imports in the test_common.py file to reflect the changes.


llucax:
After some investigation I finally found the issue. You need to also update the structure in `py/`:

* `git mv py/frequenz/api/common/metrics/ py/frequenz/api/common/v1/`
* Create a `__init__.py` in `py/frequenz/api/common/v1/`

With that the docs can be generated again.

llucax:
This is one more reminder that we should split protobuf specs from Python bindings, people working on the specs shouldn't care about how Python bindings...

tiyash-basu-frequenz:
Works now. Thanks @llucax

llucax (approved):

PR #93, opened 2023-09-26
Add new sensor category and metric enums

### LINKED ISSUE #51: Represent sensors as as their own category
### What's needed?

We have two types of equipment in our microgrids:
* electrical control or measurement component, such as inverters, batteries, meters, etc.
* sensors providing details of the surrounding environment, such as temperature, humidity, irradiance, etc.

Currently, [sensors are represented as a component category](https://github.com/frequenz-floss/frequenz-api-common/blob/v0.x.x/proto/frequenz/api/common/components.proto#L37-L38). This breaks the concept of the term "component", which is meant to refer to electrical components only.

Hence, sensors should be represented as an entity disjointed from components. We need to ensure that "sensors", as a concept,
a. exist at the same level as components
b. are not components

### Proposed solution

A simple approach to address this would be to
a. remove sensors from the `ComponentCategory` enum, AND
b. rename the `SensorType` enum to `SensorCategory`.

### Use cases

_No response_

_No response_

### Additional context

_No response_

tiyash-basu-frequenz:
In our terminology, `components` are the devices that are connected to a microgrid's electrical circuit, and `sensors` are the devices that measure the physical parameters of the environment.

Hence, we need to distinguish between the two, and we need to be able to identify the type of sensor and the type of metric that it measures.

This commit adds the `frequenz.api.common.v1.sensors` package, which contains the `SensorCategory` and `SensorMetric` enums.

Also, this commit removes
* the `COMPONENT_CATEGORY_SENSOR` variant from the `ComponentCategory` enum, and
* The `SensorType` enum.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -10,6 +10,11 @@
   `frequenz.api.common.v1.<package>`. `v1` is the API's major version, and will
   be incremented for breaking changes.
 
+- Added `frequenz.api.common.sensors` package, containing the enums
+  `SensorCategory` and `SensorType`. Removed the component category variant
+  `COMPONENT_CATEGORY_SENSOR` and the enum `SensorType` from
+  `frequenz.api.common.components`.
+
 ## New Features
 
 <!-- Here goes the main new features and examples or instructions on how to use them -->
--- a/proto/frequenz/api/common/v1/components.proto
+++ b/proto/frequenz/api/common/v1/components.proto
@@ -34,9 +34,6 @@ enum ComponentCategory {
   // A station for charging electrical vehicles.
   COMPONENT_CATEGORY_EV_CHARGER = 6;
 
-  // A sensor for measuring ambient metrics, e.g., temperature, humidity, etc.
-  COMPONENT_CATEGORY_SENSOR = 7;
-
   // A crypto miner.
   COMPONENT_CATEGORY_CRYPTO_MINER = 8;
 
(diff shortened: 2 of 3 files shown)


llucax (commented):
~LGTM, but it is missing the `py/frequenz/api/common/v1/sensors/__init__.py` (I assume this will be rebased on top of the PR adding the `v1` :) ).~ This is just adding one file, I was confused by the text in the PR description "This commit adds the `frequenz.api.common.sensors` package".

llucax (dismissed):

tiyash-basu-frequenz:
> I was confused by the text in the PR description "This commit adds the frequenz.api.common.sensors package".

Good point. Fixed.

llucax (approved):
Mmm, I still see *package* everywhere, but I'm not sure if this is even wrong in the protobuf world (in Python *packages* are like directories, that contain *modules*, which are files). So approving because I don't think it is important really.

tiyash-basu-frequenz:
I assume you mean how we declare package names in protobuf files?

Declaring `package`s is nice, because in rust it directly maps to a module. It helps provide clean and predictable namespaces. Note that [the package keyword is ignored in python](https://protobuf.dev/programming-guides/proto3/#packages:~:text=In%20Python%2C%20the%20package%20directive%20is%20ignored%2C%20since%20Python%20modules%20are%20organized%20according%20to%20their%20location%20in%20the%20file%20system.).

PR #99, opened 2023-10-04
Add a message 'Pagination' to define pagination parameters

### LINKED ISSUE #98: Add support for pagination
### What's needed?

Messages related to pagination that can be used in other APIs.

### Proposed solution

First part of proposed solution in https://github.com/frequenz-floss/frequenz-api-weather/issues/19

### Use cases

Protobuf specs for the weather api.

_No response_

### Additional context

_No response_

TalweSingh:
This message can be utilized in various protobuf specifications for APIs, allowing the request to specify pagination parameters and establish the API's response behavior.

Fixes #98

### DIFF
--- a/proto/frequenz/api/common/v1/pagination.proto
+++ b/proto/frequenz/api/common/v1/pagination.proto
@@ -1,0 +1,43 @@
+// Parameters for pagination.
+//
+// Copyright 2023 Frequenz Energy-as-a-Service GmbH
+//
+// Licensed under the MIT License (the "License");
+// you may not use this file except in compliance with the License.
+
+syntax = "proto3";
+
+package frequenz.api.common.v1.pagination;
+
+// A message defining parameters for paginating list requests.
+// It can be appended to a request message to specify the desired page of
+// results and the maximum number of results per page. For initial requests
+// (requesting the first page), the page_token should not be provided. For
+// subsequent requests (requesting any page after the first), the
+// next_page_token from the previous responses PaginationInfo must be supplied.
+// The page_size should only be specified in the initial request and will be
+// disregarded in subsequent requests.
+message PaginationParams {
+  // The maximum number of results to be returned per request.
+  optional uint32 page_size = 1;
+
+  // The token identifying a specific page of the list results.
+  optional string page_token = 2;
+}
+
+// A message providing metadata about paginated list results.
+// The PaginationInfo message delivers metadata concerning the paginated list
+// results and should be appended to the response message of a list request. The
+// total_items field must be set to the total count of items that adhere to the
+// filter criteria defined in the request. The next_page_token field should be
+// populated with the token to be used in the subsequent request to fetch the
+// next page of results. If there are no additional results, the next_page_token
+// field should be omitted.
+message PaginationInfo {
+  // The total number of items that match the filter criteria defined in the
+  // request.
+  uint32 total_items = 1;
+
+  // Token for retrieving the next set of results.
+  optional string next_page_token = 2;
+}


proto/frequenz/api/common/v1/pagination.proto
```diff
@@ -0,0 +1,34 @@
...
+  // The token identifying a specific page of the list results.
+  string page_token = 2;
+}
+
+// A message providing metadata about paginated list results.
+message PaginationInfo {
+  // [Optional]
+  // The total number of items that match the filter criteria defined in the
+  // request.
+  uint32 total_filtered_items = 1;
```
thomas-nicolai-frequenz: Is there a reason to not call this just `total_items`?

proto/frequenz/api/common/v1/pagination.proto
```diff
@@ -0,0 +1,34 @@
...
+syntax = "proto3";
+
+package frequenz.api.common.v1.pagination;
+
+// A message defining parameters for paginating list requests.
+message PaginationParams {
+  // [Optional]
+  // The maximum number of results to be returned per request.
+  // Defaults to 50 if omitted.
+  uint32 page_size = 1;
```
thomas-nicolai-frequenz: please add `optional` as a prefix. I missed that.

proto/frequenz/api/common/v1/pagination.proto
```diff
@@ -0,0 +1,34 @@
...
+
+  // [Optional]
+  // The token identifying a specific page of the list results.
+  string page_token = 2;
```
thomas-nicolai-frequenz: please add `optional` as a prefix. I missed that.

proto/frequenz/api/common/v1/pagination.proto
```diff
@@ -0,0 +1,34 @@
...
+}
+
+// A message providing metadata about paginated list results.
+message PaginationInfo {
+  // [Optional]
```
thomas-nicolai-frequenz: Please remove [Optional] as this is a response message.

proto/frequenz/api/common/v1/pagination.proto
```diff
@@ -0,0 +1,34 @@
...
+  // The total number of items that match the filter criteria defined in the
+  // request.
+  uint32 total_filtered_items = 1;
+
+  // [Optional]
```
thomas-nicolai-frequenz: Please remove [Optional] as this is a response message.

proto/frequenz/api/common/v1/pagination.proto
TalweSingh: i thought it'd be more clear to add the filtered. But it's probably clear from the context. So rather `total_items` ?

thomas-nicolai-frequenz: yeah I would keep it lean.

proto/frequenz/api/common/v1/pagination.proto
```diff
@@ -0,0 +1,34 @@
...
+// Licensed under the MIT License (the "License");
+// you may not use this file except in compliance with the License.
+
+syntax = "proto3";
+
+package frequenz.api.common.v1.pagination;
+
+// A message defining parameters for paginating list requests.
+message PaginationParams {
+  // [Optional]
```
tiyash-basu-frequenz: I do not think we need this line. Specifying it in the field definition is enough.

proto/frequenz/api/common/v1/pagination.proto
```diff
@@ -0,0 +1,34 @@
...
+  // The maximum number of results to be returned per request.
+  // Defaults to 50 if omitted.
```
tiyash-basu-frequenz: Shall we specify a default in the common specs? What if an API has no option to list 10 items per page? I feel this defaults should be left to the individual API specs.

proto/frequenz/api/common/v1/pagination.proto
```diff
@@ -0,0 +1,34 @@
...
+  uint32 page_size = 1;
+
+  // [Optional]
```
tiyash-basu-frequenz: Same. I do not think we need this line. Specifying it in the field definition is enough.

proto/frequenz/api/common/v1/pagination.proto
```diff
@@ -0,0 +1,34 @@
...
+  // The token identifying a specific page of the list results.
+  string page_token = 2;
+}
+
+// A message providing metadata about paginated list results.
```
tiyash-basu-frequenz, line 28: It would be nice if there is a code example on how to use this message. A simple RPC would be enough as an example.

proto/frequenz/api/common/v1/pagination.proto
TalweSingh: Good point, I agree.

proto/frequenz/api/common/v1/pagination.proto
TalweSingh: You mean as in how to utilize this response message to receive the next page from the API?

proto/frequenz/api/common/v1/pagination.proto
```diff
@@ -27,3 +27,42 @@ message PaginationInfo {
   // Token for retrieving the next set of results.
   string next_page_token = 2;
 }
+
+
+// !!! example "Example of a service that uses pagination."
```
TalweSingh: I think this will not be rendered on the website though. I think it's not converted to markdown by protoc-gen-doc. Any ideas on where to put it so it is rendered? 
Also I can't render the website, installed the dependencies first and the used `mkdocs-serve`. But it's not working, also no on the main branch. Does it work for you? @tiyash-basu-frequenz

tiyash-basu-frequenz: I think for now you could put code examples in backticks, like this:

```
this is a code example
```

tiyash-basu-frequenz: >Also I can't render the website, installed the dependencies first and the used `mkdocs-serve`. But it's not working, also no on the main branch. Does it work for you?

It does work for me, but I have found issues with it. Happy to talk about this and sort it out on your end in a call.

tiyash-basu-frequenz: I'd suggest adding the example to the doc for the `message`.

proto/frequenz/api/common/v1/pagination.proto
llucax: Super minor: what about `page_id`? I think it is a more common term.

thomas-nicolai-frequenz: no sure if thats correct

https://protobuf.dev/programming-guides/api/#define-pagination-api

llucax: OK, not in the protobuf world then :+1:

proto/frequenz/api/common/v1/pagination.proto
```diff
@@ -0,0 +1,43 @@
...
+}
+
+// A message providing metadata about paginated list results.
+// The PaginationInfo message delivers metadata concerning the paginated list
+// results and should be appended to the response message of a list request. The
+// total_items field must be set to the total count of items that adhere to the
+// filter criteria defined in the request. The next_page_token field should be
+// populated with the token to be used in the subsequent request to fetch the
+// next page of results. If there are no additional results, the next_page_token
+// field should be omitted.
```
tiyash-basu-frequenz, line 35: > If there are no additional results, the next_page_token field should be omitted.

Should the `next_page_token` be optional then?

proto/frequenz/api/common/v1/pagination.proto
tiyash-basu-frequenz: The word `Optional` in the docs looks incorrect since the field itself is not optional.

We should make the `page_size` field optional, right? If not specified, then it could default to service's page size (we'd need to mention this in the docs).

proto/frequenz/api/common/v1/pagination.proto
tiyash-basu-frequenz: Same. The word `Optional` in the docs looks incorrect since the field itself is not optional.

I think we want to make the `page_token` field optional? Should the default be to return the first page?

TalweSingh:
Met with Tiyash to chat about explaining the Messages better since the example isn’t rendered on the documentation website. Tiyash suggested we should put the how-to right in the docs above each message so it’s easy to find. I agree, I think it's clear like this. FYI @thomas-nicolai-frequenz.

proto/frequenz/api/common/v1/pagination.proto
thomas-nicolai-frequenz: > We should make the page_size field optional, right? If not specified, then it could default to service's page size (we'd need to mention this in the docs).

true, makes sense.

proto/frequenz/api/common/v1/pagination.proto
TalweSingh: Yes, but should we add it since it is in the response message and is handled by our API implementation? Tom suggested above to leave it out since it's in the response.

proto/frequenz/api/common/v1/pagination.proto
thomas-nicolai-frequenz: > I think we want to make the page_token field optional? Should the default be to return the first page?

two times yes.

proto/frequenz/api/common/v1/pagination.proto
tiyash-basu-frequenz: Yes, we should. Otherwise the implementations cannot return `None` when there is no next page.

thomas-nicolai-frequenz: > Should the next_page_token be optional then?

no I would say because this is used in a messages response and being set by the server. it can just be that this is empty. I mean we can add the keyword optional if that also applies to things that are part of the servers response message.

TalweSingh: Isn't every field optional in v3 of protobuf or why should this not be possible then? I think it is set to None automatically if omitted.

thomas-nicolai-frequenz:
> Met with Tiyash to chat about explaining the Messages better since the example isn’t rendered on the documentation website. 

which example?

proto/frequenz/api/common/v1/pagination.proto
tiyash-basu-frequenz: I think `None` is semantically better than an empty string to represent NULL. So I'd vote for an optional type in such cases.

tiyash-basu-frequenz: >Isn't every field optional in v3 of protobuf or why should this not be possible then?

No, scalar types are not optional by default.

TalweSingh:
There was a commit before which I removed and the example was something like this 
```
/*
* // Example of a service for listing microgrids.
*
* // Service definition containing the RPC call.
* service MicrogridService {
*   // RPC call to get a list of microgrids.
*   rpc ListMicrogrids(ListMicrogridsRequest) returns (ListMicrogridsResponse);
* }
*
* // A message representing a microgrid entity.
* message Microgrid {
*   // ID of the microgrid.
*   string microgrid_id = 1;
*   // Additional microgrid fields...
* }
*
* // A message for the ListMicrogrids RPC request.
* // It includes a filter to fetch microgrids with a specific ID.
* message ListMicrogridsRequest {
*   // Fetch microgrids with a specific ID.
*   string microgrid_id_filter = 1;
* }
*
* // A message for the ListMicrogrids RPC response.
* // It includes a list of microgrids.
* message ListMicrogridsResponse {
*   // List of microgrids.
*   repeated Microgrid microgrids = 1;
* }
*/

```

proto/frequenz/api/common/v1/pagination.proto
thomas-nicolai-frequenz: fine to follow @tiyash-basu-frequenz lead here. but lets make this clear in the documentation.

TalweSingh: <del>Ok, so i should add the prefix to the Pagination Params as well I assume? </del>
Just saw the comment above, will do!
Like this

```
...
optional page_size
...
optional page_token
...
optional next_page_token
...
```

tiyash-basu-frequenz: `page_size` and `page_token` should not be optional IMO, because these won't be null. Only the `next_page_token` field should be `optional` because it can be null.

TalweSingh: I am not 100% sure if i understand this correctly, so to clarify:
They should not be optional because it either has a value (page_token), for fetching the next request, or is omitted for the initial request. But specifically setting it to null doesn't mean anyhting in that sense, so it shouldnt be optional.
But if next_page_token is optional and omitted, it is set to null and this specifically conveys that this is the last response for the reqeust?

tiyash-basu-frequenz: When a service returns this object, then the following things will happen:

a. A page is being returned, therefore it will have a size. So `page_size` will always contains the number of items in the returned page.
b. A page is being returned, so it must have a token. So `page_token` will always representt the token of the returned page, which will never be null.
c. There may or may not be a next page. Therefore `next_page_token` is an optional field.

TalweSingh: `page_size` and `page_token` are only for the request message. So if the initial request is sent, the `page_token` should be omitted to signal this is the initial request. 
`page_size` is communicated in the initial request, so unless we want the user to have the ability to adjust the `page_size` of every response this only needs to be specified in the initial request.

thomas-nicolai-frequenz: > so unless we want the user to have the ability to adjust the page_size of every response this only needs to be specified in the initial request

I think it makes sense to offer that ability. The standard for most APIs is that the page size can be adjusted on subsequent requests.

thomas-nicolai-frequenz: Here is an example how it usually works with REST:

Initial request: GET /items?page_size=10
Response includes next_page_token="xyz"
Subsequent request: GET /items?page_size=20&next_page_token=xyz

proto/frequenz/api/common/v1/pagination.proto
tiyash-basu-frequenz:
```suggestion
  optional uint32 page_size = 1;
```

tiyash-basu-frequenz (commented):
I have just one final one from my side. Otherwise the rest looks good!

thomas-nicolai-frequenz (dismissed):

tiyash-basu-frequenz (dismissed):
Looks nice and clean! LGTM!

tiyash-basu-frequenz (approved):

PR #104, opened 2023-10-06
Make lower and upper bounds optional in the Bounds message

### LINKED ISSUE #103: Make `Bounds.lower` and `Bounds.upper` optional
### What's needed?

Users could prefer to bound a metric partially. Currently, if a user desires no upper bound for a metric, `Bounds.upper` has to be set to `f32::MAX` (pseudocoding). Similarly, if they desire no lower bound, `Bounds.lower` has to be set to `f32::MIN`.

The benefit of this approach is that you don't have to check for null or undefined conditions in your code, simplifying some of your logic.

However, setting extreme values can be less intuitive to understand when inspecting the data. Someone looking at the bounds might wonder why they are set to such extreme values and could interpret them incorrectly.

Making it optional has the following benefits:

Explicitness: Making the fields optional makes it explicit that a bound was not set, making it easier to understand when reading the data.

Flexibility: Making the bounds optional gives you more flexibility in dealing with edge cases. For example, if you later decide to handle undefined bounds differently, you can easily do so. (e.g., you could treat `None` as 0 for metrics that are always have a positive value).

Semantic Accuracy: Using optional fields for values that are truly optional is semantically more accurate.

### Proposed solution

```proto
message Bounds {
  // The lower bound.
  // If absent, there is no lower bound.
  optional float lower = 1;

  // The upper bound.
  // If absent, there is no upper bound.
  optional float upper = 2;
}
```

### Use cases

_No response_

_No response_

### Additional context

_No response_

tiyash-basu-frequenz:
The lower and upper bounds in the Bounds message are now optional. If absent, there is no lower or upper bound. This is useful for metrics that are intended to have only one bound.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,7 +6,7 @@
 
 ## Upgrading
 
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
+- `lower` and `upper` bounds fields in the `Bounds` message are now `optional`
 
 ## New Features
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,26 +2,16 @@
 
 ## Summary
 
-Added versioning to the API, added new component category variants, and split
-sensors out of components.
+<!-- Here goes a general summary of what this release is about -->
 
 ## Upgrading
 
-- The package names have been changed from `frequenz.api.common.<package>` to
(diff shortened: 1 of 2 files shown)


stefan-brus-frequenz:
These changes look like they should be mentioned in release notes

tiyash-basu-frequenz:
> look like they should be mentioned in release notes

My bad, missed it. Fixed.

stefan-brus-frequenz (approved):

PR #114, opened 2023-10-20
Update the metrics package

### LINKED ISSUE #111: Update how bounds are returned
### What's needed?

Currently system bounds are returned as a pair of inclusion and exclusion bounds. Before sending a request, clients need to check if the request parameter falls outside the exclusion bounds, and inside the inclusion bounds. Otherwise the service will return an error.

The objective is to ensure that a range of parameters are excluded from operations. E.g., we may want to exclude low charging powers for batteries.

However, since the exclusion and inclusion bounds behave in an opposite manner, clients need to write more code to satisfy the constraints. This can be improved by tweaking the way we expose these bounds.

### Proposed solution

Instead of sending exclusion and inclusion bounds, APIs could just send an array of inclusion bounds. Then clients just need to check if their request parameter satisfies all the inclusion bounds. They can reuse the same logic for all bounds, leading to less chances of error, and have a more ergonomic experience.

Here is how a definition of a simple `Metric` could look like:

```proto
// Represents a dynamic metric sample with an array of real-time system-defined bounds.
//
// !!! note "Real-time Bounds"
//     These bounds can be dynamic and may vary based on environmental conditions, 
//     state of the component, or other factors.
//
// !!! note "Interpreting Multiple Bounds"
//     When multiple bounds are specified, they are processed in the order they 
//     appear in the list. The system should define how multiple bounds interact. 
//     For example, the first bounds in the list could represent inclusion bounds, 
//     while the second could represent exclusion bounds.
//
// !!! example "Relationship Between Bounds and Metric Samples"
//     The system needs to define how it will use multiple bounds to evaluate 
//     metric samples for operations like charge and discharge commands.
//
// The diagram below illustrates the relationship between the bounds.
// ```
//   bounds[0].lower                             bounds[1].upper
// <-------|============|------------------|============|--------->
//               bounds[0].upper     bounds[1].lower
// ```
// ---- values here are disallowed and will be rejected
// ==== values here are allowed and will be accepted
message Metric {
  // The current value of the dynamic metric sample.
  float value = 1;

  // Specifies an array of dynamic bounds that apply to the metric sample.
  // These bounds may adapt in real-time to reflect the current operating conditions.
  repeated Bounds bounds = 2;
}
```

Please refer [here](https://github.com/frequenz-io/frequenz-api-reporting/issues/1) for its source and a broader context.

### Use cases

_No response_

_No response_

### Additional context

_No response_

### LINKED ISSUE #112: Remove `component_bounds`
### What's needed?

`component_bounds` contains bounds imposed by a component. However, this information is already captured in `system_inclusion_bounds`. So the former is redundant, and can be removed.

### Proposed solution

_No response_

### Use cases

_No response_

_No response_

### Additional context

_No response_

### LINKED ISSUE #113: Remove `rated_bounds`
### What's needed?

`rated_bounds` contain static limits that represent the manufacturer's ratings for a given component. Since this information is static, therefore not subject to change in a running deployment, this does not need to be streamed (`Metric` objects are intended to be streamed by gRPC server-side streaming methods).

Therefore, `rated_bounds` should be removed from metric messages, and instead be returned in messages that report static information about components, such as [this](https://github.com/frequenz-floss/frequenz-api-microgrid/blob/5e011aade569098ffd56541a4e05648869307dd9/proto/frequenz/api/microgrid/v1/microgrid.proto#L126) one..

### Proposed solution

_No response_

### Use cases

_No response_

_No response_

### Additional context

_No response_

tiyash-basu-frequenz:
This PR refactors the metrics protobuf package to simplify and generalise it further.


proto/frequenz/api/common/v1/metrics.proto
```diff
@@ -22,159 +24,222 @@ message Bounds {
...
+  oneof metric_return_type {
+    SimpleMetricSample simple_metric = 1;
+    AggregatedMetricSample aggregated_metric = 2;
+  }
+}
+
+// List of supported metrics.
+enum Metric {
+  // Default value.
+  METRIC_UNSPECIFIED = 0;
```
thomas-nicolai-frequenz, line 159: Does it make sense to have the `METRIC_` prefix? Just wondering as the enum is already called Metric.

thomas-nicolai-frequenz (dismissed):

proto/frequenz/api/common/v1/metrics.proto
tiyash-basu-frequenz: It is recommended to prefix protobuf enum variants with enum names. This is because if there are two or more enums in a scope, all their variants share the same scope (not the scope within the enum, but the outer scope to which the enums belong to).

So, we consistently prefix our protobuf enum variants with the enum names. Also, some programming languages support these by default. E.g., `METRIC_UNSPECIFIED` will appear just as `Unspecified` in rust.

thomas-nicolai-frequenz: Got it. Thanks for clarifying!

proto/frequenz/api/common/v1/components.proto
```diff
@@ -10,6 +10,217 @@ syntax = "proto3";
...
+//             "max_value": 10.6,
+//             "list_values": [9.8, 10.1, 10.6],
+//             "bounds": [
+//               { "lower": -20.0, "upper": -5.0 },
+//               { "lower": 5.0, "upper": 20.0 }
+//             ]
+//           }
+//         }
+//       },
+//       "state": "COMPONENT_STATE_CHARGING",
```
thomas-nicolai-frequenz: state is also an array now. the example needs adjusting.

proto/frequenz/api/common/v1/components.proto
```diff
@@ -10,6 +10,217 @@ syntax = "proto3";
...
+//     }
+//    ```
+message ComponentData {
+  // Representation of a sampled metric along with its value.
+  message MetricSample {
+    // The metric that was sampled.
+    frequenz.api.common.v1.metrics.Metric metric = 1;
+
+    // The value of the sampled metric.
+    frequenz.api.common.v1.metrics.MetricSampleVariant value = 2;
```
thomas-nicolai-frequenz: I would suggest to call this sample and not value.

proto/frequenz/api/common/v1/components.proto
```diff
@@ -10,6 +10,217 @@ syntax = "proto3";
...
+//       },
+//       "state": "COMPONENT_STATE_CHARGING",
+//       "errors": []
+//     }
+//    ```
+message ComponentData {
+  // Representation of a sampled metric along with its value.
+  message MetricSample {
+    // The metric that was sampled.
+    frequenz.api.common.v1.metrics.Metric metric = 1;
```
thomas-nicolai-frequenz: The prefix "frequenz.api.common.v1.metrics." can be remove as this is a self-reference.

proto/frequenz/api/common/v1/components.proto
thomas-nicolai-frequenz: And the prefix "frequenz.api.common.v1.metrics." can be remove as this is a self-reference.

proto/frequenz/api/common/v1/components.proto
```diff
@@ -10,6 +10,217 @@ syntax = "proto3";
...
+  // List of errors for the specific microgrid component.
+  repeated ComponentError errors = 4;
+}
+
+// Unified component state enum across all component categories.
+enum ComponentState {
+  // Default value.
+  COMPONENT_STATE_UNSPECIFIED = 0;
+
+  // The component is on.
```
thomas-nicolai-frequenz: Its off not on ;-)

proto/frequenz/api/common/v1/components.proto
```diff
@@ -10,6 +10,217 @@ syntax = "proto3";
...
+  COMPONENT_STATE_DISCHARGING = 7;
+
+  // The component is in an error state.
+  COMPONENT_STATE_ERROR = 8;
+
+  // The component is temporarily unavailable.
+  COMPONENT_STATE_UNAVAILABLE = 9;
+
+  // The component state is unknown.
+  COMPONENT_STATE_UNKNOWN = 10;
```
thomas-nicolai-frequenz: Is there a difference between UNKOWN and UNSPECIFIED? I feel its the same...

proto/frequenz/api/common/v1/components.proto
```diff
@@ -10,6 +10,217 @@ syntax = "proto3";
...
+}
+
+// A representation of all possible errors that can occur across all component
+// categories.
+enum ComponentError {
+  // Default value. No specific error is specified.
+  COMPONENT_ERROR_UNSPECIFIED = 0;
```
thomas-nicolai-frequenz: I would also have suggested to call this UKNOWN rather UNSPECIFIED but both is fine I guess. Just pointing it out...

proto/frequenz/api/common/v1/components.proto
```diff
@@ -10,6 +10,217 @@ syntax = "proto3";
...
+  COMPONENT_ERROR_EV_CABLE_LOCK_FAILED = 22;
+
+  // Error indicating an invalid electric vehicle (EV) cable.
+  COMPONENT_ERROR_EV_CABLE_INVALID = 23;
+
+  // Error indicating an unknown electric vehicle (EV) plug.
+  COMPONENT_ERROR_EV_PLUG_UNKNOWN = 24;
+
+  // Error indicating a non-standard electric vehicle (EV) consumer.
+  COMPONENT_ERROR_EV_CONSUMER_NON_STANDARD = 25;
```
thomas-nicolai-frequenz: I suggest to keep some gaps in the numbering between battery, ev etc.

proto/frequenz/api/common/v1/components.proto
```diff
@@ -10,6 +10,217 @@ syntax = "proto3";
...
+  COMPONENT_ERROR_ILLEGAL_COMPONENT_STATE_REQUESTED = 18;
+
+  // Error indicating that the hardware of the component is inaccessible.
+  COMPONENT_ERROR_HARDWARE_INACCESSIBLE = 19;
+
+  // Error indicating an internal error within the component.
+  COMPONENT_ERROR_INTERNAL = 20;
+
+  // Error indicating electric vehicle (EV) cable lost.
+  COMPONENT_ERROR_EV_CABLE_LOST = 21;
```
thomas-nicolai-frequenz: I would put a gap in the numbering between common errors and component category specific ones. Maybe component specific ones should start at 100 or something? I mean there is plenty of "space".

proto/frequenz/api/common/v1/metrics.proto
```diff
@@ -22,159 +24,223 @@ message Bounds {
...
+
+  // General BMS metrics.
+  METRIC_BATTERY_CAPACITY = 101;
+  METRIC_BATTERY_SOC_PCT = 102;
+  METRIC_BATTERY_SOP_CHARGE = 103;
+  METRIC_BATTERY_SOP_DISCHARGE = 104;
+  METRIC_BATTERY_TEMPERATURE = 105;
+
+  // General inverter metrics.
+  METRIC_INVERTER_TEMPERATURE = 106;
```
thomas-nicolai-frequenz: again I would leave a gap between component specific metrics.

proto/frequenz/api/common/v1/components.proto
tiyash-basu-frequenz: The prefix cannot be removed, because the definitions of `Metric` and `MetricSampleVariant` are in a different file (and therefore under a different package hierarchy).

thomas-nicolai-frequenz: oh right I overlooked that. Sorry my bad!

proto/frequenz/api/common/v1/components.proto
tiyash-basu-frequenz: There is a difference, but maybe the wording needs to be improved.

`UNSPECIFIED` is used when a sender specifies no value for the enum. Protobuf serializers/deserializers will use this as a default in such cases.

`UNKOWN` is used when we get a state from a component, but we do not know what this state means (e.g., if this is not mentioned in the manual). So this would be specified, while being unknown to us. Our friend suggests `UNDEFINED` but not sure if it is an improvement.

proto/frequenz/api/common/v1/components.proto
thomas-nicolai-frequenz: > I would suggest to call this sample and not value.

This renaming depends on wether we keep the structure as suggest or change it to guarantee key uniqueness.

proto/frequenz/api/common/v1/components.proto
thomas-nicolai-frequenz: Yeah, that makes sense. I think it would be great to add this to the UNKOWN state to explain what the difference is to the UNSPECIFIED state.

proto/frequenz/api/common/v1/components.proto
tiyash-basu-frequenz: Agree about the renaming. `sample` is more meaningful than `value` here.

tiyash-basu-frequenz:
I have addressed the review comments @thomas-nicolai-frequenz . Ready for another look.

proto/frequenz/api/common/v1/components.proto
tiyash-basu-frequenz: Tom approved verbally.

proto/frequenz/api/common/v1/components.proto
tiyash-basu-frequenz: Tom approved verbally.

proto/frequenz/api/common/v1/metrics.proto
```diff
@@ -22,159 +24,223 @@ message Bounds {
...
+  // Default value.
+  METRIC_UNSPECIFIED = 0;
+
+  // DC electricity metrics
+  METRIC_DC_VOLTAGE = 1;
+  METRIC_DC_CURRENT = 2;
+  METRIC_DC_POWER = 3;
+
+  // General AC electricity metrics
+  METRIC_AC_FREQUENCY = 4;
```
stefan-brus-frequenz: Should there be more numeric "spacing" between the enum values here? E.g. using the values 11, 12, 13, 14 for one category of things, and then having the next one start at 20, in case you later find more "things" that belong to the previous category. I see that it's already done for other enums in this PR. I am not sure if it's necessary here though since this isn't really my domain, so up to you.

stefan-brus-frequenz (dismissed):

proto/frequenz/api/common/v1/metrics.proto
tiyash-basu-frequenz: I added gaps between each group of electrical metrics.

stefan-brus-frequenz (approved):

PR #125, opened 2023-11-01
Add messages to stream sensor data

tiyash-basu-frequenz:
This PR adds a new message `SensorData`, which can used by APIs to send and receive data originating from sensors. Such data samples will contain the metric and the value reported by the sensor, and will also contain a state to inform the system if a sensor is in an error state or not.

### DIFF
--- a/proto/frequenz/api/common/v1/sensors.proto
+++ b/proto/frequenz/api/common/v1/sensors.proto
@@ -12,6 +12,8 @@ package frequenz.api.common.v1.sensors;
 
 import "frequenz/api/common/v1/metrics.proto";
 
+import "google/protobuf/timestamp.proto";
+
 // Enumerated sensor categories.
 enum SensorCategory {
   // Unspecified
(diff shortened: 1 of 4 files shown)


proto/frequenz/api/common/v1/sensors.proto
```diff
@@ -76,3 +78,102 @@ enum SensorMetric {
...
+message SensorMetricSample {
+  // The UTC timestamp of when the metric was sampled.
+  google.protobuf.Timestamp sampled_at = 1;
+
+  // The metric that was sampled.
+  frequenz.api.common.v1.metrics.Metric metric = 2;
+
+  // The value of the sampled metric.
+  frequenz.api.common.v1.metrics.MetricSampleVariant sample = 3;
+}
```
tiyash-basu-frequenz, line 181: This entire message is just a copy paste of `metrics.MetricSample`, but bounds deleted. Does it make sense to copy it like this? In the future, if we change `metrics.MetricSample`, then we also would need to change this message here. Alternatively, we could just use `metrics.MetricSample` instead of this one, but never populate the bounds for sensors.

tiyash-basu-frequenz:
@thomas-nicolai-frequenz ready for a look

proto/frequenz/api/common/v1/sensors.proto
thomas-nicolai-frequenz: > Alternatively, we could just use metrics.MetricSample instead of this one, but never populate the bounds for sensors.

No I don't feel thats a good idea.

> Does it make sense to copy it like this?

I guess so until we know more about what sensors we are going to support. My assumption is that maybe we will replace `MetricSampleVariant` with its own version just for sensors.

thomas-nicolai-frequenz (approved):

tiyash-basu-frequenz:
Rebased on top of v0.x.x.

stefan-brus-frequenz (approved):
LGTM

PR #127, opened 2023-11-03
Add messages describing microgrid entities

tiyash-basu-frequenz:
This change introduces messages to describe microgrid entities such as sensors, components, electrical connections, and microgrids.


proto/frequenz/api/common/v1/grid.proto
```diff
@@ -0,0 +1,62 @@
+// Frequenz definitions of grids as entites participating in trading.
+//
+// Copyright 2023 Frequenz Energy-as-a-Service GmbH
+//
+// Licensed under the MIT License (the "License");
+// you may not use this file except in compliance with the License.
+
+syntax = "proto3";
+
+package frequenz.api.common.v1.grid;
```
tiyash-basu-frequenz: This package name somehow looks odd. This focuses mainly on market delivery areas (and delivery periods from the Assets API), but is called `grid`.

Also, as a part of cleaning up the package names (subject to review in this PR), I added a file `components/grid.proto` to represent grid connection points in a microgrid. Having two packages named `components.grid` and `grid` could be confusing.

I have the following suggestions:
1. Rename this package, e.g. to `trade_delivery` or something similar, OR
2. Place the delivery related messages in `microgrid.proto`, because they seem to be tighly related to the microgrid domain, OR
3. Rename the package `components.grid` to `components.grid_connection_point`.

Which one would you suggest? Or something else perhaps?

proto/frequenz/api/common/v1/components.proto
```diff
@@ -75,46 +82,100 @@ enum ComponentCategory {
...
+//
+// !!! info "Historical Data"
+//     The timestamps of when a connection was created and terminated allows for
+//     tracking the changes over time to a microgrid, providing insights into
+//     when and how the microgrid infrastructure has been modified.
+//
+message ComponentConnection {
+  // Unique identifier of the component where the connection originates. This is
+  // aligned with the direction of current flow away from the grid connection
+  // point, or in case of islands, away from the islanding point.
```
tiyash-basu-frequenz, line 172: I extended the original documentation slightly, in order to accommodate islands. If this looks good, I will add this to other places `ComponentConnection`'s documentation.

proto/frequenz/api/common/v1/grid.proto
thomas-nicolai-frequenz: > Rename this package, e.g. to trade_delivery or something similar, OR

Well these things are given/provided by the grid operator and not necessary just related to trading. The assets API is also not really about trading at all. These things are really related to the grid. They only apply if a microgrid is connected to a grid. I don't have a better idea right now tbh.

> Place the delivery related messages in microgrid.proto, because they seem to be tighly related to the microgrid domain

I'm not sure. `Counterparty` of example has nothing to do with the microgrid.

> Rename the package components.grid to components.grid_connection_point.

This feels misleading. These things are like the config parameters for the grid that a microgrid is connected to. Its like the metadata of the grid and how its being operated.

proto/frequenz/api/common/v1/components/grid.proto
```diff
@@ -0,0 +1,26 @@
...
+//
+// License:
+// MIT
+
+syntax = "proto3";
+
+package frequenz.api.common.v1.components.grid;
+
+// A representation of a grid connection point.
+message Grid {
```
thomas-nicolai-frequenz: How about renaming this to `GridConnection` or `GridFuse` instead of just Grid? Grid isn't really a "component".

proto/frequenz/api/common/v1/lifetime.proto
```diff
@@ -0,0 +1,28 @@
+// Frequenz definition of entity lifetimes, based upon start and end timestamps.
+//
+// Copyright 2023 Frequenz Energy-as-a-Service GmbH
+//
+// Licensed under the MIT License (the "License");
+// you may not use this file except in compliance with the License.
+
+syntax = "proto3";
+
+package frequenz.api.common.v1.lifetime;
```
thomas-nicolai-frequenz: Why wouldn't we put that into microgrid.proto? I don't feel this should deserve its own file.

> Lifetime captures the timestamps indicating the active operational period of a microgrid asset

proto/frequenz/api/common/v1/grid.proto
thomas-nicolai-frequenz: Also if you look at this import path that looks totally legit to me. I mean its the delivery area of the grid, isn't it? It describes a certain part of the grid.

> frequenz.api.common.v1.grid.DeliveryArea

proto/frequenz/api/common/v1/lifetime.proto
tiyash-basu-frequenz: We discussed it to be in `lifetime.proto` since this is being used in several places, including `Component` and `ComponentConnection`.

proto/frequenz/api/common/v1/grid.proto
tiyash-basu-frequenz: Okay, so we leave it as is?

proto/frequenz/api/common/v1/components/grid.proto
tiyash-basu-frequenz: ChatGPT suggests that the most widely known name for this point is `Point of Common Coupling (PCC)`. However, I prefer the suggestion `Grid Interconnection Point`. What do you suggest?

image

tiyash-basu-frequenz: Or just `GridConnectionPoint` maybe?

proto/frequenz/api/common/v1/components.proto
tiyash-basu-frequenz: Added this elsewhere. Adjusted the phrases slightly to go with the flow of the original documentation.

proto/frequenz/api/common/v1/components.proto
```diff
@@ -75,46 +82,102 @@ enum ComponentCategory {
...
 }
 
-// Enumerated inverter types.
-enum InverterType {
-  // Unspecified.
-  INVERTER_TYPE_UNSPECIFIED = 0;
+// Microgrid electrical component details.
+message Component {
+  // The component ID.
+  uint64 id = 1;
```
tiyash-basu-frequenz, line 119: Forgot to mention earlier: in the original specs, the `id` field has the type `ComponentId`. Since the definition of a component ID was missing, I added the type `uint64`.

proto/frequenz/api/common/v1/lifetime.proto
thomas-nicolai-frequenz: Sure also sensors but these are all part of the microgrid. This has no application beyond microgrids. Hence the question. The most outer scope of this is the microgrid.

proto/frequenz/api/common/v1/grid.proto
thomas-nicolai-frequenz: Well I don't have a better idea right now.

tiyash-basu-frequenz: Okay, then we can keep it as for now. We can revisit later if needed.

proto/frequenz/api/common/v1/components/grid.proto
thomas-nicolai-frequenz: https://en.wikipedia.org/wiki/Microgrid#Point_of_common_coupling_(PCC)

It looks like ChatGPT is right if you look at Wikipedia. Of course GridConnectionPoint is a more general term describing the same concept without using the more sophisticated terminology. Which way we ever go we can of course also use the documentation to clarify. Why don't you put out a vote in Rnd or something?

tiyash-basu-frequenz: As discussed in slack, changed it to `GridConnectiopnPoint`. The idea is to be more approachable to users that do not have an EE background.

proto/frequenz/api/common/v1/lifetime.proto
tiyash-basu-frequenz: That actually raises a point: should we package microgrid-specific entities in a directory named `microgrid`? That would address the issue with scoping that you brought up.

Otherwise, just putting `Lifetime` in `microgrid.proto`, but leaving the rest out feels arbitrary ..

proto/frequenz/api/common/v1/components/grid.proto
thomas-nicolai-frequenz: Yes, however don't forget to mention PCC in the documentation and briefly explain why we choose grid connection point (GCP) over PCC.

proto/frequenz/api/common/v1/lifetime.proto
thomas-nicolai-frequenz: > should we package microgrid-specific entities in a directory named microgrid

we can of course but in this case we should apply the same to all the others. it also raises the question what the structure in these subfolders look like My vote would still got to put it into the microgrid.proto though. At least for now until we have established a clear structure for everything. I don't feel good moving on some parts while not moving on others.

tiyash-basu-frequenz: As discussed in-person, I updated the file structure for microgrid related entities and grid/delivery_area.

Ready for a look.

I will touch the rest (location, metrics, pagination) in a different PR.

tiyash-basu-frequenz: This is the structure we have for grid, and microgrid related files:

```
grid/
  delivery_area.proto (package: frequenz.api.common.v1.grid)
microgrid/
  components/
    battery.proto (package: frequenz.api.common.v1.microgrid.components)
    ev_charger.proto (package: frequenz.api.common.v1.microgrid.components)
    fuse.proto (package: frequenz.api.common.v1.microgrid.components)
    grid.proto (package: frequenz.api.common.v1.microgrid.components)
    inverter.proto (package: frequenz.api.common.v1.microgrid.components)
    transformer.proto (package: frequenz.api.common.v1.microgrid.components)
    components.proto (package: frequenz.api.common.v1.microgrid.components)
  sensors/
    sensors.proto (package: frequenz.api.common.v1.microgrid.sensors)
  lifetime.proto (package: frequenz.api.common.v1.microgrid)
```

So, the package of each protobuf file is also its path in the file structure.

Again, `locations.proto`, `metrics.proto`, and `pagination.proto` will be updated in a different PR.

thomas-nicolai-frequenz: I think thats reasonable. I'm still not happy with the lifetime.proto but I don't have a better idea, so lets go with that!

thomas-nicolai-frequenz (approved):

stefan-brus-frequenz (approved):

PR #131, opened 2023-11-08
Refactor protobuf messages and files to establish a standard packaging scheme

tiyash-basu-frequenz:
This change introduces the following packaging norm to the remaining protobuf files: the package for every protobuf file is it's path inside the protobuf directory. E.g., the package name for the message `PaginationInfo` is `frequenz.api.common.v1.pagination`, and the message lives in the file `frequenz/api/common/v1/paginaation/pagination_info.proto`.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -57,6 +57,8 @@
 
 - Added a message `Microgrid` to represent a microgrid.
 
+- Updated the package name of `location.proto` to `frequenz.api.common.v1`.
+
 ## New Features
 
 <!-- Here goes the main new features and examples or instructions on how to use them -->
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -59,6 +59,8 @@
 
 - Updated the package name of `location.proto` to `frequenz.api.common.v1`.
 
+- Added messages to support pagination in APIs.
+
 ## New Features
 
 <!-- Here goes the main new features and examples or instructions on how to use them -->
--- a/proto/frequenz/api/common/v1/location.proto
+++ b/proto/frequenz/api/common/v1/location.proto
@@ -7,7 +7,7 @@
 
 syntax = "proto3";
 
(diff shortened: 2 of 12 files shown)


tiyash-basu-frequenz:
Looking into the doc-generation error.

tiyash-basu-frequenz:
> Looking into the doc-generation error.

Fixed

thomas-nicolai-frequenz (approved):

llucax (approved):
LGTM. I'm curios though, what's the benefit of the split in multiple files?

stefan-brus-frequenz (approved):

tiyash-basu-frequenz:
>I'm curios though, what's the benefit of the split in multiple files?

Nothing, from the perspective of rust or python. It is just cosmetic.
Although, I feel `pagination/pagination_info.proto` and `pagination/pagination_params.proto` displays the scope better. The same applies to `bounds.proto` and `metric_sample.proto`. This is also what how the files in the google api submodule are organized.

tiyash-basu-frequenz:
> > I'm curios though, what's the benefit of the split in multiple files?
> 
> Nothing, from the perspective of rust or python. It is just cosmetic. Although, I feel `pagination/pagination_info.proto` and `pagination/pagination_params.proto` displays the scope better. The same applies to `bounds.proto` and `metric_sample.proto`. This is also what how the files in the google api submodule are organized.

Evidently, I was incorrect about python (correct about rust though). In python, the file name gets treated as the package name, so it creeps into the import paths as well. I still feel that splitting up files still adds more structure to the messages. So I would still go ahead with this, to keep things moving. If needed, we can of course look into re-organizing it before making the first major release.

image

llucax:
Personally I don't mind the extra package in Python, to be honest it already looks ugly in Python with the `_pb2` suffix and all, and in general we want to abstract all this away in the SDK anyway, so this would only affect SDK developers or people wanting to access the APIs directly.

PR #133, opened 2023-11-10
Cleanup RELEASE_NOTES.md for release v0.5.0

tiyash-basu-frequenz:


RELEASE_NOTES.md
```diff
@@ -2,7 +2,11 @@
 
 ## Summary
 
-<!-- Here goes a general summary of what this release is about -->
+This release contains
+* additions of data definitions to the common specs, that are intended to be
+  used by the Frequenz APIs.
+* refactors of existing definitions, in order to make them more consistent and
+  easier to use.
```
llucax, line 9: Just a small heads up in case you follow the release instructions, GitHub releases behaves a bit funny about Markdown in the release notes, and will keep any `\n`, so if you want to have a normal text flow, you need to avoid `\n` in the release notes.

llucax (approved):
Just a super minor styling comment, but I'm fine to moving forward as is, we can also fix the release notes in GitHub after the release, if we find it necessary.

RELEASE_NOTES.md
tiyash-basu-frequenz: Got it, thanks for the heads up.
I will go ahead with this, and later adjust the release notes in github.

PR #148, opened 2023-12-11
Add definitions for Electricity Trading API 

ghost:
Add some protobuf definitions necessary for the Electricity Trading (and for Ancillary Services Market). 

This folder structure is a proposal, if anyone is against it for whatever reason, please suggest another folder structure that you would consider better. 

FYI @thomas-nicolai-frequenz @tiyash-basu-frequenz @frequenz-floss/api-electricity-trading-team

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "submodules/api-common-protos"]
 	path = submodules/api-common-protos
-	url = https://github.com/googleapis/api-common-protos.git
+	url = https://github.com/googleapis/googleapis.git
--- a/proto/frequenz/api/common/v1/grid/delivery_duration.proto
+++ b/proto/frequenz/api/common/v1/grid/delivery_duration.proto
@@ -1,0 +1,61 @@
+// Frequenz definitions of the time increment used for electricity 
+// deliveries and trading.
+//
+// Copyright 2023 Frequenz Energy-as-a-Service GmbH
+//
+// Licensed under the MIT License (the "License");
+// you may not use this file except in compliance with the License.
+
+syntax = "proto3";
+
+package frequenz.api.common.v1.grid;
+
+import "google/protobuf/timestamp.proto";
+
+// DeliveryDuration represents the time increment, in minutes, 
+// used for electricity deliveries and trading. These durations
+// serve as the basis for defining the delivery period in contracts,
+// and they dictate how energy is scheduled and delivered to meet 
+// contractual obligations.
+//
+// !!! note "Compatibility Constraints"
+//     Not all delivery durations are universally 
+//     compatible with all delivery areas or markets.
+//
+enum DeliveryDuration {
+  // Default value, indicates that the duration is unspecified.
+  DELIVERY_DURATION_UNSPECIFIED = 0; 
+    
+  // 5-minute duration
+  DELIVERY_DURATION_5 = 1; 
+
+  // 15-minute contract duration.
+  DELIVERY_DURATION_15 = 2; 
+  
+  // 30-minute contract duration.
+  DELIVERY_DURATION_30 = 3; 
+  
+  // 1-hour contract duration.
+  DELIVERY_DURATION_60 = 4; 
+}
+
+// DeliveryPeriod represents the time period during which the contract 
+// is delivered. It is defined by a start timestamp and a duration.
+message DeliveryPeriod {
+  // Start UTC timestamp represents the beginning of the delivery period.
+  // This timestamp is inclusive, meaning that the delivery period starts
+  // from this point in time.
+  // 
+  // !!! note
+  //     Delivery period start time constraints:
+  //     - 5-minute durations must start at times that are multiples of 
+  //       5 minutes past the hour.
+  //     - 15-minute durations must start at :00, :15, :30, or 
+  //       :45 past the hour.
+  //     - 30-minute durations must start at :00 or :30 past the hour.
+  //     - 60-minute durations must start at :00 past the hour.
+  google.protobuf.Timestamp start = 1;
+
+  // The length of the delivery period.
+  DeliveryDuration duration = 2;
+}
--- a/proto/frequenz/api/common/v1/market/price.proto
+++ b/proto/frequenz/api/common/v1/market/price.proto
@@ -1,0 +1,53 @@
+// Frequenz definitions of price for electricity trading.
+//
+// Copyright 2023 Frequenz Energy-as-a-Service GmbH
+//
+// Licensed under the MIT License (the "License");
+// you may not use this file except in compliance with the License.
+
+syntax = "proto3";
+
+package frequenz.api.common.v1.market;
+
+import "google/type/decimal.proto";
+
+// Represents a monetary price for electricity trading, including
+// the amount and currency. The currency used should align with the
+// delivery area's standard currency.
+//
+// !!! caution "Validation Required"
+//     It's essential to ensure that the currency aligns with the 
+//     standard currency of the associated delivery area. Failure to
+//     do so may result in the API service rejecting the request due to currency
+//     mismatches.
+//
+// !!! info "Relation to Delivery Area"
+//     The currency specified is intrinsically related to the delivery area 
+//     for the contract. Make sure the chosen currency is compatible with 
+//     the delivery area's currency standards.
+//
+message Price {
+  // List of supported currencies.
+  // 
+  // !!! info "Extensibility"
+  //     New currencies can be added to this enum to support additional markets.
+  enum Currency {
+    CURRENCY_UNSPECIFIED = 0;
+    CURRENCY_USD = 1;
+    CURRENCY_CAD = 2;
+    CURRENCY_EUR = 3;
+    CURRENCY_GBP = 4;
+    CURRENCY_CHF = 5;
+    CURRENCY_CNY = 6;
+    CURRENCY_JPY = 7;
+    CURRENCY_AUD = 8;
+    CURRENCY_NZD = 9;
+    CURRENCY_SGD = 10;
+  }
+
+  // The amount of the price.
+  google.type.Decimal amount = 1; 
+
+  // The currency in which the price is denominated.
+  Currency currency = 2;
+}
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -16,7 +16,7 @@ This release:
 
 ## New Features
 
-<!-- Here goes the main new features and examples or instructions on how to use them -->
+Add protobufs definition necesary for Electricity Trading (and for Ancillary Services Market). 
 
 ## Bug Fixes
 
--- a/py/frequenz/api/common/v1/market/__init__.py
+++ b/py/frequenz/api/common/v1/market/__init__.py
@@ -1,0 +1,4 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""Frequenz common gRPC API and bindings."""
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,8 @@ classifiers = [
 ]
 requires-python = ">= 3.11, < 4"
 dependencies = [
-  "googleapis-common-protos >= 1.56.2, < 2",
+  "googleapis-common-protos >= 1.56.4, < 2",
+  "googleapis-common-protos-stubs == 2.2.0",
   "grpcio >= 1.51.1, < 2",
 ]
 dynamic = ["version"]


llucax:
I guess there was already some design discussion around this, but I will mention it just in case. I wonder if it would be better to split this repo (or start new ones) based on functionality. Like having a `api-common-migrogrid`, `api-common-trading`, otherwise people only wanting to use common stuff for the microgrid will need to pull common code for all APIs, which could be a bit overkill.

Also in the future it could be harder to maintain, because if, for example, you need to make a breaking change in the `DeliveryDuration`, we'll need to create a new major release that will affect also the `api-microgrid` (and each and every API out there), when the breaking change doesn't affect it at all, and probably only affect a handful of APIs.

thomas-nicolai-frequenz:
> Like having a api-common-migrogrid, api-common-trading, otherwise people only wanting to use common stuff for the microgrid

@llucax this would lead to a hell lot of duplicated code e.g. microgrid and reporting API. they are almost using the same. Also with regards to the Platform Assets API, this API is (re-)using almost everything from all the other APIs. Splitting this all up will also lead to a lot of inconsistencies between the various APIs over time.

thomas-nicolai-frequenz:
> you need to make a breaking change in the DeliveryDuration, we'll need to create a new major release that will affect also the api-microgrid (and each and every API out there), when the breaking change doesn't affect it at all, and probably only affect a handful of APIs.

Yes and its good like that because otherwise you do not break it on the protobuf level but on the logical level. These things have to stay aligned!

thomas-nicolai-frequenz (dismissed):

proto/frequenz/api/common/v1/market/price.proto
```diff
@@ -0,0 +1,52 @@
...
+    CURRENCY_SGD = 10;
+    // ...
+  }
+
+  // The amount of the price.
+  google.type.Decimal amount = 1; 
+
+  // The currency in which the price is denominated.
+  Currency currency = 2;
+}
```
tiyash-basu-frequenz: Missing newline :)

proto/frequenz/api/common/v1/market/price.proto
```diff
@@ -0,0 +1,52 @@
...
+    CURRENCY_CAD = 2;
+    CURRENCY_EUR = 3;
+    CURRENCY_GBP = 4;
+    CURRENCY_CHF = 5;
+    CURRENCY_CNY = 6;
+    CURRENCY_JPY = 7;
+    CURRENCY_AUD = 8;
+    CURRENCY_NZD = 9;
+    CURRENCY_SGD = 10;
+    // ...
```
tiyash-basu-frequenz: Unnecessary comment :)

proto/frequenz/api/common/v1/grid/delivery_duration.proto
```diff
@@ -0,0 +1,59 @@
...
+  //       5 minutes past the hour.
+  //     - 15-minute durations must start at :00, :15, :30, or 
+  //       :45 past the hour.
+  //     - 30-minute durations must start at :00 or :30 past the hour.
+  //     - 60-minute durations must start at :00 past the hour.
+  google.protobuf.Timestamp start = 1;
+
+  // Duration represents the length of the delivery period in minutes.
+  DeliveryDuration duration = 2;
+}
```
tiyash-basu-frequenz: Missing newline :)

proto/frequenz/api/common/v1/grid/delivery_duration.proto
```diff
@@ -0,0 +1,59 @@
...
+  //     - 5-minute durations must start at times that are multiples of 
+  //       5 minutes past the hour.
+  //     - 15-minute durations must start at :00, :15, :30, or 
+  //       :45 past the hour.
+  //     - 30-minute durations must start at :00 or :30 past the hour.
+  //     - 60-minute durations must start at :00 past the hour.
+  google.protobuf.Timestamp start = 1;
+
+  // Duration represents the length of the delivery period in minutes.
+  DeliveryDuration duration = 2;
```
tiyash-basu-frequenz: `... the delivery period in minutes.`

I would just say the `... the delivery period.`, and get rid of `in minutes.`. Because whether it is minutes or not, is actually controlled in the `DeliveryDuration` enum definition, and not here.

proto/frequenz/api/common/v1/grid/delivery_duration.proto
```diff
@@ -0,0 +1,59 @@
...
+// deliveries and trading.
+//
+// Copyright 2023 Frequenz Energy-as-a-Service GmbH
+//
+// Licensed under the MIT License (the "License");
+// you may not use this file except in compliance with the License.
+
+syntax = "proto3";
+
+package frequenz.api.common.v1.grid;
```
tiyash-basu-frequenz, line 11: Is this import being used anywhere?

proto/frequenz/api/common/v1/market/price.proto
thomas-nicolai-frequenz: good catch!

tiyash-basu-frequenz:
>  Like having a `api-common-migrogrid`, `api-common-trading`, otherwise people only wanting to use common stuff for the microgrid will need to pull common code for all APIs, which could be a bit overkill.

Why not import just the relevant subdirectories from the common repo? Also, is the overhead of including all the common specs too much? It is (as of now) a relatively small protobuf library, and the packaging structure should help users understand the scopes of its protobuf messages well enough.

Also, hypothetically speaking, if we ever head towards splitting up this repo, I'd be in favour of adding definitions to `frequenz-api-microgrid` instead of creating `api-common-microgrid`.

proto/frequenz/api/common/v1/grid/delivery_duration.proto
tiyash-basu-frequenz: Ah sorry, please ignore this comment. I somehow misread `package` as `import` :face_with_spiral_eyes:

proto/frequenz/api/common/v1/market/price.proto
```diff
@@ -0,0 +1,52 @@
+// Frequenz definitions of price for electricity trading.
+//
+// Copyright 2023 Frequenz Energy-as-a-Service GmbH
+//
+// Licensed under the MIT License (the "License");
+// you may not use this file except in compliance with the License.
+
+syntax = "proto3";
```
tiyash-basu-frequenz, line 8: This file needs a package definition. Looking at the file structure, I think it would be
```proto
package frequenz.api.common.v1.market;
```

thomas-nicolai-frequenz:
> Also, hypothetically speaking, if we ever head towards splitting up this repo, I'd be in favour of adding definitions to frequenz-api-microgrid instead of creating api-common-microgrid.

lol very good point!

tiyash-basu-frequenz:
@camille-bouvy-frequenz since we are adding a new package here, I'd recommend adding `market/____init__.py` to this directory: https://github.com/frequenz-floss/frequenz-api-common/tree/v0.x.x/py/frequenz/api/common/v1

ghost:
@thomas-nicolai-frequenz I get an error in the packages import because `google/type/decimal.proto` is not found. It seems like is not explicitly listed as part of the `googleapis-common-protos` package. Should we then directly include it in the repo?

tiyash-basu-frequenz:
@camille-bouvy-frequenz you need to do the following to get access to the `Decimal` type:
1. Remove the submodule `api-common-protos`
2. Add this repo as a submodule: https://github.com/googleapis/googleapis, but please use the name `api-common-protos`, to keep this as a non-breaking change for now. I will then create a separate issue to rename this submodule and add it to the next major release milestone.

llucax:
> @llucax this would lead to a hell lot of duplicated code e.g. microgrid and reporting API. they are almost using the same. Also with regards to the Platform Assets API, this API is (re-)using almost everything from all the other APIs. Splitting this all up will also lead to a lot of inconsistencies between the various APIs over time.

I didn't mean to duplicate stuff, but have a more hierarchical dependency tree, so `api-trading-common` would use `api-microgrid-common`, and both a `api-common` for example. Anyway, maybe it's too soon and a flat structure can help us move faster at first, but I think this will eventually need to happen.

proto/frequenz/api/common/v1/grid/delivery_duration.proto
llucax: While we are at it I would also remove the redundancy of "Duration represents" and just go for "The length of the delivery period.".

llucax:
> Also, hypothetically speaking, if we ever head towards splitting up this repo, I'd be in favour of adding definitions to `frequenz-api-microgrid` instead of creating `api-common-microgrid`.

Maybe, the issue with that is if you just want to import a message for `Component`, you'll end up pulling the whole gRPC interface too, which is overkill. Maybe at some point we need to split gRPC interfaces from protobuf messages.

But yeah, the more I think about it, the more I believe it is too soon to split stuff beyond the protobuf/python split.

ghost:
@tiyash-basu-frequenz Ok thanks this worked! I am still getting a mypy error caused by missing type stubs for the `google.type.decimal_pb2` module. I looked it up a bit so far and the only solution I was able to find so far is to manually create them. Do you know of a better fix?

proto/frequenz/api/common/v1/grid/delivery_duration.proto
```diff
@@ -0,0 +1,61 @@
+// Frequenz definitions of the time increment used for electricity 
+// deliveries and trading.
```
thomas-nicolai-frequenz, line 2: I would avoid the term `deliveries`!

tiyash-basu-frequenz:
>I am still getting a mypy error caused by missing type stubs for the google.type.decimal_pb2 module. I looked it up a bit so far and the only solution I was able to find so far is to manually create them. Do you know of a better fix?

I looked into it briefly, and I could not find a solution that uses existing stubs. Perhaps it would be better idea to add an exception for this specific type for now, to unblock this PR.

@llucax any thoughts?

llucax:
Having a look now.

.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "submodules/api-common-protos"]
 	path = submodules/api-common-protos
-	url = https://github.com/googleapis/api-common-protos.git
+	url = https://github.com/googleapis/googleapis.git
```
llucax, line 3: So, this change of URL can't be done without also changing the Python dependencies, because we are using a package providing the built `api-common-protos`.

We need to find the package that provides the whole `googleapis` repo instead.

llucax (commented):
I just checked out the new submodule and it is huge. I wonder if we really want to pull the whole definition of all Google APIs just to get a decimal type, it seems a bit overkill.

Can't we just copy&paste that singular proto definition here or does google also adds support functions to all languages to handle it?

Maybe even then we can implement it in this repo too, at least in python I guess it's just instantiating a  [`Decimal`](https://docs.python.org/3/library/decimal.html#decimal-objects) object with the protobuf message info.

.gitmodules
llucax: This is where the dependency is declared: https://github.com/frequenz-floss/frequenz-api-common/blob/v0.x.x/pyproject.toml#L29

llucax:
I'm finding too many issues with the switch of submodule, for example in the new submodule I can't find any `timestamp.proto` file, which we are currently using.

llucax:
OK, now I am really confused. Some observations to discuss (I put them there so I don't forget).

* `google/protobuf/timestamp.proto` doesn't live in [googleapis/api-common-protos] either. It seems to live in the [protobuf] Python package, so I guess it is part of the base protobuf definitions, even when it is under the `google/` *namespace*.

* We are only using `timestamp.proto` from the google namespace up to this PR, so I'm not sure why we were including the submodule in the past.

* Type checking with `mypy` worked so far because we are using the [mypy-protobuf] package (provided indirectly by `repo-config`).

* We are using the [googleapis-common-protos] python package to provide defnitions of [googleapis/api-common-protos]. This package is provided by Google.

* [googleapis-common-protos] doesn't provide any type information for mypy.

* There is a [googleapis-common-protos-stubs] Python package providing the mypy typing info but it's **not** done by Google and [it doesn't look super *healthy* according to Snyk](https://snyk.io/advisor/python/googleapis-common-protos-stubs) [^1].

* `google/types/decimal.proto` is present in [googleapis/googleapis] but not in [googleapis/api-common-protos].

* The contents of the [googleapis-common-protos] is weird. It includes `google/types/decimal.proto` which is not included in [googleapis/api-common-protos], but it doesn't nearly include everything in [googleapis/googleapis] either.

So the situation looks pretty ugly. I can see a few of paths forward:

1. Leave things are they are now and just add the [googleapis-common-protos-stubs] package to the dependencies here (`googleapis-common-protos-stubs == 2.2.0`).
2. Build all python bindings from google protobufs ourselves (in this or a separate repo), including mypy type info.
3. Remove all dependencies from Google and copy `decimal.proto` to this repo.

Option 1 is probably the simplest and more straightforward, but we are not making this mess any better, if any, maybe a bit worse because we keep adding dependencies, and one that is not official and we don't know how well maintained it is.

Option 2 seems a bit overkill and a lot of work.

Option 3 seems to be the best IMHO, it will simplify a lot this repo and it shouldn't be much work.

If we want to unblock this PR, I'd suggest option 1 for now, I just checked and `nox` passes all tests with this diff applied:

```diff
diff --git a/pyproject.toml b/pyproject.toml
index 681d206..ffb44bb 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,2 +56,3 @@ dev-mypy = [
   "mypy == 1.7.1",
+  "googleapis-common-protos-stubs == 2.2.0",
   "grpc-stubs == 1.53.0.3",
```

[googleapis/api-common-protos]: https://github.com/googleapis/api-common-protos/
[googleapis/googleapis]: https://github.com/googleapis/googleapis
[protobuf]: https://pypi.org/project/protobuf/
[mypy-protobuf]: https://pypi.org/project/mypy-protobuf/
[googleapis-common-protos]: https://pypi.org/project/googleapis-common-protos/
[googleapis-common-protos-stubs]: https://pypi.org/project/googleapis-common-protos-stubs/

[^1]: Not many users and not updated often, but that could only be because there is still not a very wide adoption of type hints in Python and because the proto definitions don't change ofter, so there is no need for updates. But we might consider to build our own if we need to.

ghost:
> google/protobuf/timestamp.proto doesn't live in [googleapis/api-common-protos](https://github.com/googleapis/api-common-protos/) either.

Indeed, you don't need a submodule for `google.protobuf` types (explained in more details [here](https://github.com/googleapis/api-common-protos))

tiyash-basu-frequenz:
Thanks @llucax 
I think option 1 is the path of least resistance for now.

thomas-nicolai-frequenz:
One more option would be to replace the decimal type with our own definition:

```
message Decimal {
  int64 integer_part = 1;
  int32 scale = 2;
}
```

ghost:
As we discussed the 1st option will be implemented for now (i.e adding the [googleapis-common-protos-stubs](https://pypi.org/project/googleapis-common-protos-stubs/) package to the dependencies) so that it can be merged. Some alternatives to having the `decimal.proto` file from the `googleapis/googleapis` submodule will be discussed in another issue.

tiyash-basu-frequenz:
@camille-bouvy-frequenz you just need to rebase, and this should be good to go!

llucax:
> I think option 1 is the path of least resistance for now.

Yup, I opened a couple of issues to be addressed in the future:

* #151 
* #152

tiyash-basu-frequenz (approved):
LGTM!

PR #149, opened 2023-12-13
Add Max charge and discharge values for a component.

thea-leake:
Components were missing rated charge and discharge values.

Rated bounds had been removed from ComponentData, as this is not changing,
and serves different use case than normal Bounds.

* Adds `ComponentRatedBound` message to associate a charge & discharge range
  to a specific `Metric`.
* Adds rated bounds field to `Component` for charge & discharge
  ranges for relevant metrics.

### DIFF
--- a/proto/frequenz/api/common/v1/microgrid/components/components.proto
+++ b/proto/frequenz/api/common/v1/microgrid/components/components.proto
@@ -10,6 +10,7 @@ syntax = "proto3";
 
 package frequenz.api.common.v1.microgrid.components;
 
+import "frequenz/api/common/v1/metrics/bounds.proto";
 import "frequenz/api/common/v1/metrics/metric_sample.proto";
 import "frequenz/api/common/v1/microgrid/components/battery.proto";
 import "frequenz/api/common/v1/microgrid/components/ev_charger.proto";
@@ -141,6 +142,26 @@ message Component {
 
   // The operational lifetime of the component.
   frequenz.api.common.v1.microgrid.Lifetime operational_lifetime = 9;
+
+  // List of rated bounds present for the component identified by Metric.
+  repeated MetricConfigBounds metric_config_bounds = 10;
+}
+
+// MetricConfigBounds describes a set of limits for a specific metric consisting
+// of a lower and upper bound for said metric.
+//
+// This can be used for example to specify an allowed range of power output
+// for a component.
+message MetricConfigBounds {
+  // Metric type the config bounds are for
+  frequenz.api.common.v1.metrics.Metric metric = 1;
+
+  // The set of bounds for the specified metric.
+  //
+  // This contains the lower and upper bounds for said metric.
+  // Sources these may be derived from include the component configuration,
+  // manufacturers limits, and limits of other devices.
+  frequenz.api.common.v1.metrics.Bounds config_bounds = 2;
 }
 
 // ComponentConnection describes a single electrical link between two components
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,6 +4,10 @@
 This release:
 * Corrects `SensorData` and `ComponentData` doc examples
   to correctly reflect differences in respective values.
+* Adds `ComponentRatedBound` message to associate a charge & discharge range
+  to a specific `Metric`.
+* Adds rated bounds field to `Component` for charge & discharge
+  ranges for relevant metrics.
 
 
 ## Upgrading


proto/frequenz/api/common/v1/microgrid/components/components.proto
```diff
@@ -141,8 +142,29 @@ message Component {
 
   // The operational lifetime of the component.
   frequenz.api.common.v1.microgrid.Lifetime operational_lifetime = 9;
+
+  // List of rated bounds present for the component identified by Metric.
+  repeated ComponentRatedBound rated_bounds = 10;
 }
 
+// Defines a components rated bound for a specific metric.
+message ComponentRatedBound {
```
tiyash-basu-frequenz: I'd rather call this `MetricRatedBounds` since these are rated bounds for a specific metric.

proto/frequenz/api/common/v1/microgrid/components/components.proto
```diff
@@ -141,8 +142,29 @@ message Component {
 
   // The operational lifetime of the component.
   frequenz.api.common.v1.microgrid.Lifetime operational_lifetime = 9;
+
+  // List of rated bounds present for the component identified by Metric.
+  repeated ComponentRatedBound rated_bounds = 10;
 }
 
+// Defines a components rated bound for a specific metric.
```
tiyash-basu-frequenz:
```suggestion
// Defines rated bounds for a specific metric.
```

proto/frequenz/api/common/v1/microgrid/components/components.proto
```diff
@@ -141,8 +142,29 @@ message Component {
...
+  // Metric type the rated bounds are for
+  frequenz.api.common.v1.metrics.Metric metric = 1;
+
+  // The maximum discharge and charge values for the device as
+  // rated by the manufacturer.
+  //
+  // Max discharge is contained in `lower`.
+  // Max charge is contained in `upper`.
+  //
+  // Max discharge (`lower`) is negative, as per passive sign convention.
```
tiyash-basu-frequenz: I'd suggest making the documentation more general, something along the lines of the following:

"The rated bounds of the specified metric. These indicate the lower and upper bounds of a metric, according to the component manufacturer's ratings."

tiyash-basu-frequenz:
@thomas-nicolai-frequenz FYI, this PR extends the `Component` message to add rated bounds for component metrics. This is a non-breaking change though.

proto/frequenz/api/common/v1/microgrid/components/components.proto
thea-leake: Updated

proto/frequenz/api/common/v1/microgrid/components/components.proto
thea-leake: Updated

proto/frequenz/api/common/v1/microgrid/components/components.proto
thea-leake: Agreed, updated :)

proto/frequenz/api/common/v1/microgrid/components/components.proto
```diff
@@ -141,8 +142,27 @@ message Component {
...
+// Defines rated bound for a specific metric.
+message MetricRatedBounds {
+  // Metric type the rated bounds are for
+  frequenz.api.common.v1.metrics.Metric metric = 1;
+
+  // The rated bounds of the specified metric.
+  //
+  // This contains the lower and upper bounds
+  // for said metric as specified by the component
+  // manufacturer's ratings.
```
tiyash-basu-frequenz: I'd suggest going for 80 character widths here. So, if a word makes the line cross the 80 character mark, then we make a new line from that word onwards. It would make the comments look more consistent.

proto/frequenz/api/common/v1/microgrid/components/components.proto
```diff
@@ -141,8 +142,27 @@ message Component {
...
+  frequenz.api.common.v1.metrics.Bounds rated_bounds = 2;
 }
 
+
+
+
```
tiyash-basu-frequenz: Looks like extra spaces. I'd suggest removing them.

tiyash-basu-frequenz (commented):
Looks good overall!

proto/frequenz/api/common/v1/microgrid/components/components.proto
thea-leake: Sounds good, done

proto/frequenz/api/common/v1/microgrid/components/components.proto
thea-leake: done

proto/frequenz/api/common/v1/microgrid/components/components.proto
```diff
@@ -141,6 +142,21 @@ message Component {
...
+// Defines rated bound for a specific metric.
+message MetricRatedBounds {
+  // Metric type the rated bounds are for
+  frequenz.api.common.v1.metrics.Metric metric = 1;
+
+  // The rated bounds of the specified metric.
+  //
+  // This contains the lower and upper bounds for said metric as specified
+  // by the component manufacturer's ratings.
+  frequenz.api.common.v1.metrics.Bounds rated_bounds = 2;
```
tiyash-basu-frequenz: After a brief discussion with @thomas-nicolai-frequenz , it is now clearer that this is not just the manufacturer's rated bounds, but a more general pair of static bounds. What this means is that these bounds can be populated by combining multiple sources.

E.g., for min/max AC power for inverters, two inputs would be required:
a. The fuse limits of the inverter
b. The manufacturer limits of the inverter
The final static bounds would then be the intersection of these two inputs.

So, `rated_bounds` is not an accurate name for this. Perhaps `config_bounds`/`MetricConfigBounds`?
Any other alternatives?

thea-leake: I like `config_bounds` & `MetricConfigBounds`, maybe `static_bounds` & `MetricStaticBounds` might be appropriate too so long as we can count on not these not changing without some event that we would restart the core apps for.  `config_bounds` works for me though.

thomas-nicolai-frequenz: I still feel we need to properly explain/document what a "MetricBound" is. Its more that these bounds limit something like power output and it happens that power is expressed in a certain metric. Metrics itself don't have bounds. You know what I mean?

tiyash-basu-frequenz: > I like `config_bounds` & `MetricConfigBounds`, maybe `static_bounds` & `MetricStaticBounds` might be appropriate too so long as we can count on not these not changing without some event that we would restart the core apps for. `config_bounds` works for me though.

I have a slight preference for `config_bounds`, over `static_bounds`, because it indicates that these bounds come from user-configs, and are subject to change when the configs change.

tiyash-basu-frequenz: >I still feel we need to properly explain/document what a "MetricBound" is. Its more that these bounds limit something like power output and it happens that power is expressed in a certain metric. Metrics itself don't have bounds. You know what I mean?

We will extend the documentation to make it clearer. Would `ComponentConfigBounds` be more suitable name?

thea-leake: Sounds good, I've updated the name to use `config_bounds` & `MetricConfigBounds`, and updated the documentation for `MetricConfigBound`

thea-leake: > Would ComponentConfigBounds be more suitable name?

I think `MetricConfigBounds` might make a bit more sense, since these are only being used in `Component`.

Having component bounds within the component wouldn't describe what the bounds are for in relation to the component; whereas `MetricConfigBounds` would be a bit more descriptive in that regard.
If we extend the usage beyond just `Component` the name would no longer be descriptive.

That being said I'm Ok with either, I think they both would work in this context.

tiyash-basu-frequenz: Then `MetricConfigBounds` it is :)

tiyash-basu-frequenz (approved):
Looks nice! LGTM!

PR #150, opened 2023-12-13
Fix doc examples for `SensorData` and `ComponentData`

thea-leake:
`SensorData` docs currently reflect components/`ComponentData`, and `ComponentData` states were missing `warnings` field contained in just `ComponentState`.

PR corrects contained examples in documentation, they now correctly reflect their respective components.


tiyash-basu-frequenz (approved):
LGTM

PR #161, opened 2023-12-27
Add Energy message for electricity trading markets

matthias-wende-frequenz:
This energy type quantizes energy in MWh and is used for api that deal with electricity trading markets.


proto/frequenz/api/common/v1/market/energy.proto
```diff
@@ -0,0 +1,24 @@
+// Frequenz definitions of price for electricity trading.
```
tiyash-basu-frequenz: Should this line be changed to something more suitable?

tiyash-basu-frequenz (approved):
LGTM!

PR #179, opened 2024-01-30
Add backwards compatibility with v0.3.x

llucax:
This PR includes the top-level files present in the v0.3.x release, so the Microgrid API v0.15.x can depend on a newer version of the Common API.

When the Microgrid API v0.15.x is phased out we can remove all the files related to v0.3.x from this repository.

### DIFF
--- a/proto/frequenz/api/common/components.proto
+++ b/proto/frequenz/api/common/components.proto
@@ -1,0 +1,132 @@
+// Frequenz microgrid components definitions.
+//
+// Copyright:
+// Copyright 2023 Frequenz Energy-as-a-Service GmbH
+//
+// License:
+// MIT
+
+syntax = "proto3";
+
+package frequenz.api.common.components;
+
+// Enumrated component categories.
+enum ComponentCategory {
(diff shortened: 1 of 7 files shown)


proto/frequenz/api/common/components.proto
```diff
@@ -0,0 +1,132 @@
+// Frequenz microgrid components definitions.
+//
+// Copyright:
+// Copyright 2023 Frequenz Energy-as-a-Service GmbH
```
tiyash-basu-frequenz, line 4: Should we put 2024 as the year?

llucax: I don't think so, this file was created in 2023 originally.

llucax:
Enabled auto-merge.

TalweSingh (approved):

llucax:
@tiyash-basu-frequenz we proceeded because you gave your blessing before, you only posted one small comment, and we want to fix this dependency issue ASAP, but if you have any further objections we can address them in a separate PR/release.

PR #181, opened 2024-02-02
Add field `source` to `MetricSample`

### LINKED ISSUE #172: Add support to annotate same metrics from different sources for a single component
### What's needed?

Components can sometimes report the same metric, but from different sensors in different places. E.g., a hybrid inverter can report DC voltage from a battery string and a PV string, or an inverter with multiple DC strings in general.

This is currently not supported in our metric definitions. We have means to define [simple and aggregated metrics samples](https://github.com/frequenz-floss/frequenz-api-common/blob/7997835d103221125258632173cab6de8fd697cd/proto/frequenz/api/common/v1/metrics/metric_sample.proto#L17-L46), and means to [associate metric samples with metric names](https://github.com/frequenz-floss/frequenz-api-common/blob/7997835d103221125258632173cab6de8fd697cd/proto/frequenz/api/common/v1/metrics/metric_sample.proto#L167-L204), but no way of annotating the sensor type (the DC string name, in case inverters have several of these).

We need means to annotate the sensor type somehow, or it would be impossible to encode metric sources in some cases (again in the case of inverters with multiple DC strings).

### Idea 1: Add a new field to `MetricSample` (non-breaking)

We could add a new field to the [`MetricSample`](https://github.com/frequenz-floss/frequenz-api-common/blob/7997835d103221125258632173cab6de8fd697cd/proto/frequenz/api/common/v1/metrics/metric_sample.proto#L167) message, which would contain the additional information about the sensor. Perhaps just an optional string would be fine here.

### Idea 2: Add a new sample variant (non-breaking)

Like `SimpleMetricSample` and `AggregatedMetricSample`, we could add a new message that contains the sensor annotation.

### Use cases

_No response_

_No response_

### Additional context

_No response_

tiyash-basu-frequenz:
This commit adds an optional string fieldthat can be used to identify the source of the metric. This is useful in cases where the same metric can be obtained from multiple sensors or sources in the same component.

E.g., a hybrid inverter can have  multiple DC strings, each with its own current and voltage sensors. In such cases, the source field can be used to differentiate between the different sources.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -10,7 +10,9 @@
 
 ## New Features
 
-<!-- Here goes the main new features and examples or instructions on how to use them -->
+- Added a field named `source` to `MetricSample` to allow the user to identify
+  the source of the metric, in case different sensors in component report
+  metrics with the same `Metric` variant.
 
 ## Bug Fixes
 
--- a/proto/frequenz/api/common/v1/metrics/metric_sample.proto
+++ b/proto/frequenz/api/common/v1/metrics/metric_sample.proto
@@ -201,4 +201,20 @@ message MetricSample {
   // ---- values here are disallowed and will be rejected
   // ==== values here are allowed and will be accepted
   repeated Bounds bounds = 4;
+
+  // An optional string that can be used to identify the source of the metric.
+  //
+  // This is expected to be populated when the same `Metric` variant can be
+  // obtained from multiple sensors in the component. Knowing the source of the
+  // metric can help in certain control and monitoring applications.
+  //
+  // E.g., a hybrid inverter can have a DC string for a battery and another DC
+  // string for a PV array. The source names could resemble, say,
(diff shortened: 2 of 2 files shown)


llucax (commented):
The PR and commit descriptions are still mentioning the word `tag`. Other than that, LGTM.

tiyash-basu-frequenz:
> The PR and commit descriptions are still mentioning the word `tag`. Other than that, LGTM.

Updated

stefan-brus-frequenz (dismissed):
LGTM

RELEASE_NOTES.md
```diff
@@ -10,7 +10,9 @@
 
 ## New Features
 
-<!-- Here goes the main new features and examples or instructions on how to use them -->
+- Added a field named `source` to `MetricSample` to allow the user to specify
```
cwasicki: "specify" -> "identify"?

cwasicki (dismissed):

cwasicki (approved):

PR #187, opened 2024-02-16
Remove dependency on `googleapis`

### LINKED ISSUE #152: Depending on `googleapis/googleapis` is overkill
### What happened?

[googleapis/googleapis](https://github.com/googleapis/googleapis) is about 67MiB, and we only use one file, `decimal.proto`, which is just encapsulating a string.

### What did you expect instead?

Having something more lighweight, like our own decimal message.

### Affected part(s)

Build script, CI, dependencies, etc. (part:tooling), The protocol buffer definition files (part:protobuf)

### Extra information

*  This probably needs some investigation to see if how different target languages deal with these `Decimal` types, and if we can get away with it. In Python for example it is trivial to construct a [`decimal.Decimal`](https://docs.python.org/3/library/decimal.html#decimal-objects) from the string.

* Origin of this issue: https://github.com/frequenz-floss/frequenz-api-common/pull/148#issuecomment-1857712305

llucax:
To be able to do so we create a new `Decimal` message that is compatible with Google's `Decimal` message.

Fixes #152.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,11 +25,7 @@ classifiers = [
   "Typing :: Typed",
 ]
 requires-python = ">= 3.11, < 4"
-dependencies = [
-  "googleapis-common-protos >= 1.56.4, < 2",
-  "googleapis-common-protos-stubs == 2.2.0",
-  "grpcio >= 1.51.1, < 2",
-]
+dependencies = ["grpcio >= 1.51.1, < 2", "protobuf >= 4.25.3, < 5"]
 dynamic = ["version"]
 
 [[project.authors]]
@@ -59,10 +55,7 @@ dev-mypy = [
   # For checking the noxfile, docs/ script, and tests
   "frequenz-api-common[dev-mkdocs,dev-noxfile,dev-pytest]",
 ]
-dev-noxfile = [
-  "nox == 2023.4.22",
-  "frequenz-repo-config[api] == 0.8.0",
-]
+dev-noxfile = ["nox == 2023.4.22", "frequenz-repo-config[api] == 0.8.0"]
 dev-pylint = [
   "pylint == 3.0.3",
   # For checking the noxfile, docs/ script, and tests
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "submodules/api-common-protos"]
-	path = submodules/api-common-protos
-	url = https://github.com/googleapis/googleapis.git
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,11 +6,20 @@
 
 ## Upgrading
 
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
+- The dependency on `googleapis-common-protos` / `googleapis/googleapis` was
+  removed, now the built-in `frequenz.api.common.v1.types.Decimal` is used
+  internally instead. This is compatible with Google's type, but if you
+  depended indirectly on Google's submodule or python packages, you may need to
+  update your dependencies.
+
+  Nevertheless it is strongly recommended to remove the dependency on Google
+  repos and use the built-in `frequenz.api.common.v1.types.Decimal` instead if
+  your project only uses the `Decimal` type from it too.
 
 ## New Features
 
-<!-- Here goes the main new features and examples or instructions on how to use them -->
+- Added a `Frequenz.api.common.v1.types.Decimal` type, compatible with
+  `google.type.Decimal`.
 
 ## Bug Fixes
 
--- a/proto/frequenz/api/common/v1/market/energy.proto
+++ b/proto/frequenz/api/common/v1/market/energy.proto
@@ -9,7 +9,7 @@ syntax = "proto3";
 
 package frequenz.api.common.v1.market;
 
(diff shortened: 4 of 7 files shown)


pyproject.toml
```diff
@@ -25,11 +25,7 @@ classifiers = [
   "Typing :: Typed",
 ]
 requires-python = ">= 3.11, < 4"
-dependencies = [
-  "googleapis-common-protos >= 1.56.4, < 2",
-  "googleapis-common-protos-stubs == 2.2.0",
-  "grpcio >= 1.51.1, < 2",
-]
+dependencies = ["grpcio >= 1.51.1, < 2", "protobuf >= 4.25.3, < 5"]
```
cwasicki, line 28: Why is this `protobuf` dependency added here?

llucax: Because we still use many `google.protobuf.xxx` messages that are included in the `protoc` compiler but still need the python part of it, before that dependency was indirectly pulled by `googleapis-common-protos`.

cwasicki (dismissed):

proto/frequenz/api/common/v1/decimal.proto
```diff
@@ -0,0 +1,25 @@
+// An arbitrary precision real number.
+//
+// Copyright 2024 Frequenz Energy-as-a-Service GmbH
+//
+// Licensed under the MIT License (the "License");
+// you may not use this file except in compliance with the License.
+
+syntax = "proto3";
+
+package frequenz.api.common.decimal;
```
tiyash-basu-frequenz:
```suggestion
package frequenz.api.common.v1;
```

tiyash-basu-frequenz:
Since this would a breaking change, marking it blocked until we release milestone [v0.5.5](https://github.com/frequenz-floss/frequenz-api-common/milestone/13).

proto/frequenz/api/common/v1/decimal.proto
llucax: Ah, good catch! Updated.

llucax:
> Since this would a breaking change, marking it blocked until we release milestone [v0.5.5](https://github.com/frequenz-floss/frequenz-api-common/milestone/13).

We can also branch v0.5.x now and merge this to v0.x.x, but I'm fine with the approach that's the easiest for you.

llucax:
What I forgot to do, since this is a breaking, is to create a `v2/` directory already to avoid ending up with the dependency issue we had with 0.5.x and previous versions.

Also fixing the package name, I noticed a whole package for decimal might be too much and maybe we can make like google and have a *types* package, for example.

I wanted to know your opinion before I update the PR.

tiyash-basu-frequenz:
>We can also branch v0.5.x now and merge this to v0.x.x, but I'm fine with the approach that's the easiest for you.

Thanks for being patient :pray: I will get v0.5.5 out today. That would be easiest for me. We do not have upcoming non-breaking changes, so we can unblock this PR later today.

> What I forgot to do, since this is a breaking, is to create a v2/ directory already to avoid ending up with the dependency issue we had with 0.5.x and previous versions.

I'd suggest holding on to that idea until we have a stable v1.0. I think the upcoming release v0.6.0 is going to be our stable-release candidate.

> I noticed a whole package for decimal might be too much and maybe we can make like google and have a types package, for example.

I would prefer that as well. Also, we'd need to create a new directory `v1/types`, and put the `decimal.proto` file inside it. (Also, as of now, the `decimal.proto` file is in the `v1` directory, but is in the package `v1.decimal`, which is not according to our standards.)

llucax:
> I'd suggest holding on to that idea until we have a stable v1.0. I think the upcoming release v0.6.0 is going to be our stable-release candidate.

Still, this will make it very messy to update again, we'll have to update everything at the same time or we won't be able to use python projects with mixed versions.

My learning is, we should have probably called pre-release versions `v0a1/`, `v0a2/`, ..., etc. or something like that, so we can create new directories without feeling guilty of taking `v1/` as a development version.

If this is really the only breaking change in v0.6.x, I guess it shouldn't be that bad, because I guess downstream projects can still depend on googleapis on their side. The wire format didn't really change.

llucax:
> I would prefer that as well. Also, we'd need to create a new directory `v1/types`, and put the `decimal.proto` file inside it. 

Yeah. that's exactly what I had in mind, will do that.

> (Also, as of now, the `decimal.proto` file is in the `v1` directory, but is in the package `v1.decimal`, which is not according to our standards.)

Ah, right, I thought I copied what `location` did, but I guess I got it wrong. Anyway, will put in the `types` package.

llucax:
Updated!

tiyash-basu-frequenz:
>Still, this will make it very messy to update again, we'll have to update everything at the same time or we won't be able to > use python projects with mixed versions.
> My learning is, we should have probably called pre-release versions

Yeah, I feel so too. Missed opportunity.

>If this is really the only breaking change in v0.6.x

No, there are quite a few in the milestone.

tiyash-basu-frequenz:
@llucax looks like a few git conflicts need resolving.
Also, this PR is technically unblocked now, but you may want to wait for rebasing until this is merged: https://github.com/frequenz-floss/frequenz-api-common/pull/192

llucax:
@tiyash-basu-frequenz done!

tiyash-basu-frequenz (approved):
LGTM

PR #188, opened 2024-02-16
Add metrics variants for inverter temperatures

### LINKED ISSUE #144: Add more temperature metrics
### What's needed?

Some inverters provide multiple temperatures. The largest set we have encountered so far consists of the following: cabinet temperature, heatsink temperature, transformer temperature.

Currently we have only one inverter temperature metrics: [`InverterTemperature`](https://github.com/frequenz-floss/frequenz-api-common/blob/dbbf62e818f409875661ea05e9371d6fde06dac6/proto/frequenz/api/common/v1/metrics/metric_sample.proto#L140), which is used to represent cabinet temperatures (which would be the general temperature inside the inverter, and not specific to components like heat-sinks or transformers).

We need to extend the [`Metrics`](https://github.com/frequenz-floss/frequenz-api-common/blob/dbbf62e818f409875661ea05e9371d6fde06dac6/proto/frequenz/api/common/v1/metrics/metric_sample.proto#L66) enum to contain more temperature metrics.

### Proposed solution

Instead of adding inverter-specific metrics, I would propose adding general temperature metrics, like `METRIC_TEMPERATURE_HEATSINK`, and `METRIC_TEMPERATURE_TRANSFORMERS`.

Alternatively, we could add a new variant in `MetricSampleVariant`, that allows us to report several values of the same metric, but tagged with a sensor name.

### Use cases

_No response_

_No response_

### Additional context

_No response_

tiyash-basu-frequenz:


cwasicki (approved):

thomas-nicolai-frequenz (approved):
Looks good.

PR #208, opened 2024-03-22
Remove `APPARENT_` from current metric names

### LINKED ISSUE #178: Replace `AC_APPARENT_CURRENT` with `AC_CURRENT`
### What's needed?

In AC (alternating current) electricity, when you multiply current (I) with voltage (V), you get what is known as apparent power. Apparent power is measured in volt-amperes (VA) and is the vector-sum of active power (measured in watts, W) and reactive power (measured in volt-amperes reactive, VAR).

This is the reason why we have used [`AC_APPARENT_CURRENT`](https://github.com/frequenz-floss/frequenz-api-common/blob/7997835d103221125258632173cab6de8fd697cd/proto/frequenz/api/common/v1/metrics/metric_sample.proto#L81-L84) instead of `AC_CURRENT` so far.

However, the more colloquial term would be `AC_CURRENT`. We should consider renaming `AC_APPARENT_CURRENT` to `AC_CURRENT`.

<details>
  <summary> Here are a few words of wisdom from ChatGPT</summary>

When designing APIs for software developers, especially in the context of energy-tech where precision and clarity are crucial, the term you choose should accurately reflect the kind of data or measurement the API is providing. 

- If your API is providing data about the current in an AC (Alternating Current) system without any reference to the voltage or power factor, then the term `ac_current` would be appropriate. This term would typically represent just the current measurement (in amperes, A) in the AC system.

- However, if your API is providing data that combines both current and voltage measurements, and is specifically related to apparent power (which in AC systems is a product of current, voltage, and the cosine of the phase angle between them), then a term like `ac_apparent_power` would be more appropriate. It's important to note that apparent power is measured in volt-amperes (VA), not in amperes, so the term `ac_apparent_current` would be a misnomer. 

In summary, choose `ac_current` for current measurements and `ac_apparent_power` for measurements related to apparent power. Remember, clarity and precision in terminology are key in API design, as they directly impact the ease of use and understanding for developers interfacing with your system.
</details>

### Proposed solution

_No response_

### Use cases

_No response_

_No response_

### Additional context

_No response_

tiyash-basu-frequenz:
Current metrics have been renamed:
- `METRIC_AC_APPARENT_CURRENT` -> `METRIC_AC_CURRENT`
- `METRIC_AC_APPARENT_CURRENT_PHASE_[1|2|3]` -> `METRIC_AC_CURRENT_PHASE_[1|2|3]`

Closes #178


nhatcher-frequenz (approved):
LGTM!

PR #209, opened 2024-03-22
Unabbreviate `THD` in the Metric enum

### LINKED ISSUE #177: Un-abbreviate `THD` in the `Metric` enum
### What's needed?

`THD` (Total Harmonic Distortion) is present as an abbreviation in the [`Metric`](https://github.com/frequenz-floss/frequenz-api-common/blob/7997835d103221125258632173cab6de8fd697cd/proto/frequenz/api/common/v1/metrics/metric_sample.proto#L128-L132) enum here:

https://github.com/frequenz-floss/frequenz-api-common/blob/7997835d103221125258632173cab6de8fd697cd/proto/frequenz/api/common/v1/metrics/metric_sample.proto#L128-L132

There are t wo possible approaches here:

1. We decided early on that we want to limit the use of abbreviations as much as possible. From that principle, we should replace `THD` with `TOTAL_HARMONIC_DISTORTION`.

2. We do use some popular abbreviations, such as `AC` and `DC`. If we consider `THD` as a popular abbreviation, then we should leave it as is.

<details>
  <summary> Here are a few words of wisdom from ChatGPT</summary>

Abbreviating "Total Harmonic Distortion" as "THD" is quite common and generally considered a good idea, especially in professional and technical contexts where the term is well understood. Here are a few reasons why abbreviating it is beneficial:

1. **Industry Standard**: THD is a standard abbreviation in the fields of electrical engineering and power systems. Professionals in these fields are familiar with the term, so using the abbreviation can facilitate clearer and more efficient communication.

2. **Conciseness**: Using "THD" makes communication more concise, especially in technical documentation, reports, or discussions where the term may be used frequently.

3. **Clarity**: In technical writing and discussion, using well-known abbreviations can actually enhance clarity by avoiding the repetition of lengthy terms.

However, it's important to consider the audience. If you're communicating with an audience that might not be familiar with the term (like non-technical staff, clients outside the engineering field, or in educational materials for beginners), it's a good practice to use the full term "Total Harmonic Distortion" at least once before switching to the abbreviation "THD." This approach ensures understanding and avoids confusion. Given your background in a technical field, using "THD" in your professional environment is likely appropriate and understood.
</details>

### Proposed solution

_No response_

### Use cases

_No response_

_No response_

### Additional context

_No response_

tiyash-basu-frequenz:
Metrics for Total Harmonic Distortion have been un-abbreviated:
- `METRIC_AC_THD_CURRENT` -> `METRIC_AC_TOTAL_HARMONIC_DISTORTION_CURRENT`
- `METRIC_AC_THD_CURRENT_PHASE_[1|2|3]` -> `METRIC_AC_TOTAL_HARMONIC_DISTORTION_CURRENT_PHASE_[1|2|3]`

Closes #177


nhatcher-frequenz (approved):
LGTM!

PR #223, opened 2024-04-30
Make Dependabot check for updates monthly on Thursdays

llucax:


llucax:
Trivial, I will force-merge to reduce the dependabot noise.

PR #232, opened 2024-07-31
Change `PaginationParams` to make `page_size` and `page_token` mutually exclusive

Marenz:
This is a suggestion on how we could make the pagination params less error-prone.
This has not been actively discussed yet.

### DIFF
--- a/proto/frequenz/api/common/v1/pagination/pagination_params.proto
+++ b/proto/frequenz/api/common/v1/pagination/pagination_params.proto
@@ -11,16 +11,21 @@ package frequenz.api.common.v1.pagination;
 
 // A message defining parameters for paginating list requests.
 // It can be appended to a request message to specify the desired page of
-// results and the maximum number of results per page. For initial requests
-// (requesting the first page), the page_token should not be provided. For
-// subsequent requests (requesting any page after the first), the
-// next_page_token from the previous responses PaginationInfo must be supplied.
-// The page_size should only be specified in the initial request and will be
-// disregarded in subsequent requests.
+// results and the maximum number of results per page.
 message PaginationParams {
-  // The maximum number of results to be returned per request.
-  optional uint32 page_size = 1;
+  oneof params {
+    // The maximum number of results to return in a single page. The service may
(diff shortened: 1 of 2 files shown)


proto/frequenz/api/common/v1/pagination/pagination_params.proto
```diff
@@ -11,16 +11,18 @@ package frequenz.api.common.v1.pagination;
...
+    // return fewer results than requested. If unspecified, the service will
+    // choose a reasonable default.
+    // May only be specified in the first request.
+    uint32 page_size = 1;
 
-  // The token identifying a specific page of the list results.
-  optional string page_token = 2;
+    // A token identifying a page of results to return. This should be the value
+    // of the `next_page_token` field in the previous response's PaginationInfo.
+    // For the first request, this field should be omitted.
```
stefan-brus-frequenz, line 25: Maybe add an example noting that the service that implements this could return an object containing the page size here, to avoid keeping it in the server state?

proto/frequenz/api/common/v1/pagination/pagination_params.proto
```diff
@@ -11,16 +11,18 @@ package frequenz.api.common.v1.pagination;
...
-// next_page_token from the previous responses PaginationInfo must be supplied.
-// The page_size should only be specified in the initial request and will be
-// disregarded in subsequent requests.
+// results and the maximum number of results per page.
 message PaginationParams {
-  // The maximum number of results to be returned per request.
-  optional uint32 page_size = 1;
+  oneof params {
+    // The maximum number of results to return in a single page. The service may
+    // return fewer results than requested. If unspecified, the service will
```
TalweSingh: I would leave the default behaviour up to each service and not specify here that the service will choose a reasonable default.

Marenz: You mean, whether its optional or not should be up to the service?

TalweSingh: Yes

stefan-brus-frequenz (approved):
LGTM

PR #235, opened 2024-08-06
Add energy metric information

flora-hofmann-frequenz:
This is the same additional info as in the [common-client](https://github.com/frequenz-floss/frequenz-client-common-python/pull/39/files).

### DIFF
--- a/proto/frequenz/api/common/v1/metrics/metric_sample.proto
+++ b/proto/frequenz/api/common/v1/metrics/metric_sample.proto
@@ -63,6 +63,20 @@ message MetricValueVariant {
 }
 
 // List of supported metrics.
+//
+// !!! note
+// AC energy metrics information:
+//
+// * This energy metric is reported directly from the component, and not a
+//  result of aggregations in our systems. If a component does not have this
+//  metric, this field cannot be populated.
+//
+//  * Components that provide energy metrics reset this metric from time to
+//  time. This behaviour is specific to each component model. E.g., some
+//  components reset it on UTC 00:00:00.
+//
+//  * This energy metric does not specify the timestamp since when the energy
+//  was being accumulated, and therefore can be inconsistent.
 enum Metric {
   // Default value.
   METRIC_UNSPECIFIED = 0;
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -8,3 +8,5 @@ This release adds a new component category `COMPONENT_CATEGORY_HVAC` to the API.
 
 - A new component category `COMPONENT_CATEGORY_HVAC` has been added to the API
   to represent HVAC (Heating, Ventilation, and Air Conditioning) systems.
+
+- Additional information for energy metric


proto/frequenz/api/common/v1/metrics/metric_sample.proto
```diff
@@ -63,6 +63,19 @@ message MetricValueVariant {
...
+// * This energy metric is reported directly from the component, and not a
+//  result of aggregations in our systems. If a component does not have this
+//  metric, this field cannot be populated.
+//
+//  * Components that provide energy metrics reset this metric from time to
+//  time. This behaviour is specific to each component model. E.g., some
+//  components reset it on UTC 00:00:00.
+//
+//  * This energy metric does not specify the timestamp since when the energy
+//  was being accumulated, and therefore can be inconsistent.
```
tiyash-basu-frequenz: LGTM in general. Should we make it more visible by making it a [note](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Howto/Markdown_in_MDN#notes_warnings_and_callouts)?

That way it looks more like a notice, and not general documentation about this enum.

flora-hofmann-frequenz: A bunch of CI test are failing now that I was trying to mark it down as a note. Should I change it back?

proto/frequenz/api/common/v1/metrics/metric_sample.proto
```diff
@@ -63,6 +63,17 @@ message MetricValueVariant {
...
+> [!NOTE]
+> AC energy metrics information:
+> * This energy metric is reported directly from the component, and not a
+>  result of aggregations in our systems. If a component does not have this
+>  metric, this field cannot be populated.
+>  * Components that provide energy metrics reset this metric from time to
+>  time. This behaviour is specific to each component model. E.g., some
+>  components reset it on UTC 00:00:00.
+>  * This energy metric does not specify the timestamp since when the energy
+>  was being accumulated, and therefore can be inconsistent.
```
tiyash-basu-frequenz: The comment looks incorrect .. these should start wist `//`, no?

Maybe better to use the MkDocs format here. Here is an example:

```
// !!! note
//     This is a note.
//     This is the second line.
```

flora-hofmann-frequenz: Ah sorry, I thought it was suppose to look like in the markup docs (but maybe I looked at the wrong one).

llucax: MkDocs now supports also GitHub format toob(via a plugin, not sure if this repo is up to date enough to use it though). in any case, using GitHub format probably makes sense only in markdown files that will actually be interpreted by GitHub, which is not the case for `.proto` files, so I agree it might be better to use MkDocs format instead here.

flora-hofmann-frequenz: Sounds good :-)

tiyash-basu-frequenz (approved):
LGTM!

PR #240, opened 2024-09-10
Use the mimimum supported `protobuf` and `grpcio` to build the project

llucax:
- **Be more explicit about the constraints on the dependencies**
- **Use the minimum dependencies for building the project**


llucax:
This approach was more complicated and didn't work.

PR #241, opened 2024-09-10
Pin grpc related dependencies

llucax:
We can't use wide dependencies because otherwise when building the wheel, the latest version is used, but the generate code should use the minimum supported version for the generation, not the latest one. See: https://protobuf.dev/support/cross-version-runtime-guarantee/

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,17 +1,5 @@
 # Frequenz Common API Release Notes
 
-## Summary
-
-<!-- Here goes a general summary of what this release is about -->
-
-## Upgrading
-
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
-
-## New Features
-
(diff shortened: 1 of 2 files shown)


RELEASE_NOTES.md
```diff
@@ -1,10 +1,5 @@
...
-## Summary
+## Bug Fixes
 
-This release adds a new component category `COMPONENT_CATEGORY_HVAC` to the API.
-
-## New Features
-
-- A new component category `COMPONENT_CATEGORY_HVAC` has been added to the API
-  to represent HVAC (Heating, Ventilation, and Air Conditioning) systems.
+- Fix a dependency issue by pinning the `grpcio-tools` version.
```
daniel-zullo-frequenz: Maybe something like:
```suggestion
- Fix a dependency issue by pinning the `grpcio` version and related libraries.
```

daniel-zullo-frequenz (dismissed):
Just a minor optional suggestion to check for. LGTM

PR #252, opened 2024-09-12
Move `conftest.py` to the python package root

llucax:
This is mostly to make it easier to navigate to the root of the package.

We also fix its location in `MANIFEST.in` to ensure it is excluded from the binary distribution.


tiyash-basu-frequenz:
Does this need a release note? LGTM otherwise

tiyash-basu-frequenz (approved):

llucax:
Nah, this is just internal (or fix a bug in the distribution package, but I think it's mostly noise, I don't think any user will care). Skipping release notes.

PR #259, opened 2024-09-27
Improve the `Lifetime` message documentation

### LINKED ISSUE #250: Add more specifications for `Lifetime` fields
### What's needed?

The `Lifetime` fields need to be better documented. E.g., one field [here](https://github.com/frequenz-floss/frequenz-api-common/blob/a76abc36888ddf14977d36dddb805506f7071bc1/proto/frequenz/api/common/v1/microgrid/components/components.proto#L146-L147) is not well documented. It does not specify what should be the behaviour if
- the `operational_lifetime` field is set to `None`, or
- if it is present but both start and stop fields in the `Lifetime` message are empty.

### Proposed solution

We need to define the requirement here. Here are two proposals:

## Proposal 1

Entities which have lifetimes set to `None` are invalid.

## Proposal 2

Entities which have lifetimes set to `None` are valid, but only if at least one of the `start_timestamp` and `end_timestamp` fields are specified.

For all the above proposals, if only the `start_timestamp` is provided, then the component is valid and active. If the `end_timestamp` is provided, the component is considered valid, but inactive past the end timestamp.

### Use cases

_No response_

_No response_

### Additional context

_No response_

llucax:
Make the specification of what `Lifetime` message is used for as clear and precise as possible.

Fixes #250.

### DIFF
--- a/proto/frequenz/api/common/v1/microgrid/lifetime.proto
+++ b/proto/frequenz/api/common/v1/microgrid/lifetime.proto
@@ -15,14 +15,27 @@ import "google/protobuf/timestamp.proto";
 // a microgrid asset, such as a component, connection, sensor, or any other
 // entity with a limited operational lifetime.
 //
-// !!! warning "Permanent Deletion"
-//     The `end_timestamp` indicates that the asset has been permanently removed
-//     from the system.
+// It is normally used to track the historical evolution of assets, and it
+// determines the validity period for the asset's data.
 //
+// If the `start_timestamp` is not set, the asset is considered to be in
+// operation since the beginning of the system's operational history.
+//
+// If the `end_timestamp` is not set, the asset is considered to be in
+// operation indefinitely into the future.
+//
+// If the Lifetime message is completely missing, it means both timestamps are
+// not set, which means the asset is considered to be in operation since the
+// beginning of the system's operational history and indefinitely into the
+// future.
 message Lifetime {
   // The timestamp when the asset became operationally active.
+  // If not set, the asset is considered to be in operation since the beginning
+  // of the system's operational history.
   google.protobuf.Timestamp start_timestamp = 1;
 
   // Optional timestamp when the asset's operational activity ceased.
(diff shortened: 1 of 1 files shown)


proto/frequenz/api/common/v1/microgrid/lifetime.proto
```diff
@@ -15,14 +15,27 @@ import "google/protobuf/timestamp.proto";
...
-//     The `end_timestamp` indicates that the asset has been permanently removed
-//     from the system.
+// It is normally used to track the historical evolution of assets, and it
+// determines the validity period for the asset's data.
 //
+// If the `start_timestamp` is not set, the asset is considered to be in
+// operation since the beginning of the system's operational history.
+//
+// If the `end_timestamp` is not set, the asset is considered to be in
+// operation until the end of the system's operational history.
```
tiyash-basu-frequenz: The "end of history" part reads a bit odd, because history indicates past tense. I feel that phrase does not add anything here, and can be removed:
```suggestion
// If the `end_timestamp` is not set, the asset is considered to be in
// operation.
```

The same comment applies to other instances where end-timestamp is being documented.

llucax:
Skipping release notes, I think they are not necessary for such a small documentation change.

proto/frequenz/api/common/v1/microgrid/lifetime.proto
llucax: What about "... is considered to be in operation indefinitely in the future"? I would like to make it explicit that it is in operation now but also at any time in the future.

tiyash-basu-frequenz: Sounds good. Should then be "indefinitely _into_ the future".

llucax: Updated.

llucax:
Enabling auto-merge.

tiyash-basu-frequenz (approved):
LGTM!

PR #291, opened 2025-03-27
Rename microgrid.components -> microgrid.electrical_components

tar-viturawong-frequenz:
This change is in preparation for communication components to just the protobuf definition. The PR disambiguates electrical components by adding `Electrical` prefix to message/enum types, `electrical_` to module and path names, and `ELECTRICAL_` to enum fields.

I have not extensively updated the comments in the protobuf definitions a that would have created an unwieldy diff that obscures the substantive change. I can follow up on that afterwards.

PR #292, opened 2025-03-27
Add protobuf definition for communication components

tar-viturawong-frequenz:


proto/frequenz/api/common/v1/microgrid/communication_components/communication_components.proto
llucax, line 1: Nitpick: I think we normally use `proto/frequenz/api/common/v1/microgrid/communication_components.proto` if it is only one file (i.e. no subdirectory).

llucax (commented):
Not sure if this was discussed anywhere, but I find it confusing that **communication components** are called **components** but have nothing to do with the existing [`Component`](https://github.com/frequenz-floss/frequenz-api-common/blob/9a71b3c74efd2466482c8ff3853580f03154842c/proto/frequenz/api/common/v1/microgrid/components/components.proto#L120). Can we name them differently to avoid confusion?

Does this mean the microgrid API will get a new RPC for listing and getting these new type of devices as we have for components and sensors? A bit more context of where this is coming from would be useful.

llucax:
Oh, I see https://github.com/frequenz-floss/frequenz-api-common/pull/291 now.

tiyash-basu-frequenz:
Thanks @tar-viturawong-frequenz. Could you please add an entry to the RELEASE_NOTES file as well?

tar-viturawong-frequenz:
> but I find it confusing that communication components are called components but have nothing to do with the existing [Component](https://github.com/frequenz-floss/frequenz-api-common/blob/9a71b3c74efd2466482c8ff3853580f03154842c/proto/frequenz/api/common/v1/microgrid/components/components.proto#L120). Can we name them differently to avoid confusion?

@llucax Unfortunately this is not a decision that I can change on my own. The term `communication component` was decided by the team some time ago, even if informally, and been incorporated into the UI as such. Renaming decision would have to come from upstream. When/If that happens, I'm happy to submit another change.

llucax:
> @llucax Unfortunately this is not a decision that I can change on my own. The term communication component was decided by the team some time ago, even if informally, and been incorporated into the UI as such. Renaming decision would have to come from upstream. When/If that happens, I'm happy to submit another change.

Yeah, all good, I didn't see that we were renaming component -> electrical_component when I wrote that, and wasn't aware this was coming up, this is why I was a bit surprised. But with the rename, it make sense.

llucax (commented):
LGTM but leaving the approval to @tiyash-basu-frequenz

proto/frequenz/api/common/v1/microgrid/communication_components.proto
```diff
@@ -0,0 +1,71 @@
+// Frequenz microgrid communication components definitions.
+//
+// Copyright:
+// Copyright 2023 Frequenz Energy-as-a-Service GmbH
```
tiyash-basu-frequenz: Date needs to be 2025.

proto/frequenz/api/common/v1/microgrid/communication_components.proto
```diff
@@ -0,0 +1,71 @@
...
+
+package frequenz.api.common.v1.microgrid.communication_components;
+
+
+// CommunicationComponent represents a communication component within a
+// microgrid.
+//
+// Communication components are networked devices responsible for enabling
+// communication between electrical components, sensors, controllers, and other
+// networked equipment.
```
tiyash-basu-frequenz: Nit: I _think_ the term is `network devices`, but I could be wrong.

proto/frequenz/api/common/v1/microgrid/communication_components.proto
```diff
@@ -0,0 +1,71 @@
...
+//
+// License:
+// MIT
+
+syntax = "proto3";
+
+package frequenz.api.common.v1.microgrid.communication_components;
+
+
+// CommunicationComponent represents a communication component within a
```
tiyash-basu-frequenz: Nit: I would recommend putting the name of the struct in backticks, so that it gets formatted accordingly in the auto-generated docs.

proto/frequenz/api/common/v1/microgrid/communication_components.proto
```diff
@@ -0,0 +1,71 @@
...
+  string name = 3;
+
+  // Category identifying the type of the communication component.
+  CommunicationComponentCategory category = 4;
+
+  // List of IP addresses assigned to this communication component.
+  repeated string ip_addresses = 5;
+}
+
+// CommunicationComponentCategory enumerates possible types of communication
```
tiyash-basu-frequenz: Nit: I would recommend putting the name of the struct in backticks, so that it gets formatted accordingly in the auto-generated docs.

tiyash-basu-frequenz (commented):
Looks great. I just have a few nits and one date related comment.

pytests/test_common.py
```diff
@@ -181,6 +181,14 @@ def test_module_import_microgrid_components() -> None:
     assert transformer_pb2_grpc is not None
 
 
+def test_module_import_microgrid_communication_components() -> None:
+    """Test that the modules can be imported."""
+    # pylint: disable=import-outside-toplevel
+    from frequenz.api.common.v1.microgrid import communication_components
```
llucax, line 195: This should be `communication_components_pb2`.

tar-viturawong-frequenz: Fixed here https://github.com/frequenz-floss/frequenz-api-common/pull/292/commits/167052f4c90cf3a8e381e8b23e796f4bae2064c9

tiyash-basu-frequenz (approved):

PR #307, opened 2025-04-08
Add enum `ElectricalComponentControlMode`

### LINKED ISSUE #264: Confusing Naming of Component State Enums: `ComponentStatus` and `ComponentStateCode`
### What's needed?

We currently have two component state enums:

- [`ComponentStatus`](https://github.com/frequenz-floss/frequenz-api-common/blob/3d0aa691af167f135801ffe58584612127ece6d3/proto/frequenz/api/common/v1/microgrid/components/components.proto#L101-L117): This enum depicts the state of our interaction with a component (e.g., whether we can read from or write to it).
- [`ComponentStateCode`](https://github.com/frequenz-floss/frequenz-api-common/blob/3d0aa691af167f135801ffe58584612127ece6d3/proto/frequenz/api/common/v1/microgrid/components/components.proto#L259C6-L259C24): This enum represents the actual state of the component as derived from the component directly.

The current naming can potentially lead to confusion between `ComponentStatus` and `ComponentStateCode`. They could be mistaken for referring to the same entity. Furthermore, `ComponentStatus` lacks specificity, as it seems to indicate the component's overall status rather than focusing on the aspect of interaction. 

We need to address the naming issue to improve readability.

### Proposed solution

<details>
<summary>Proposal 1</summary>
<br>
We can rename the `ComponentStatus` enum to something more appropriate.

One proposal is `ComponentAccessLevel`. This correctly depicts the purpose of this entity.
</details>

<details>
<summary>Proposal 2</summary>
<br>

We flip the concept, and do the following:

**`Active` mode**

As soon as a component moves to this mode, we try to start it up. Note that this cannot be guaranteed to succeed, since a component could refuse to start up, e.g., due to the component's internal configuration.

In this mode, clients are allowed to read data from the component, and send commands to it. Sending commands to a component in this mode can be prevented by setting its respective bounds to (0,0), which would make the component read-only.

**`Disabled` mode**

As soon as a component moves to this mode, we try to stop it. Again, this cannot be guaranteed to succeed due the same reasons mentioned above.

In this mode, clients do not receive any data from the component, and cannot send any control commands to it.

In this concept, `RequestedComponentMode` may be a better name.

</details>

<details>
<summary>Proposal 3</summary>
<br>

This extends Proposal 2, with an intermediate mode for switching the component off while still getting data from it.

**`Active` mode**

As soon as a component moves to this mode, we try to start it up. Note that this cannot be guaranteed to succeed, since a component could refuse to start up, e.g., due to the component's internal configuration.

In this mode, clients are allowed to read data from the component, and send commands to it. Sending commands to a component in this mode can be prevented by setting its respective bounds to (0,0), which would make the component read-only.

**`Off` mode**

As soon as a component moves to this mode, we try to stop it. Again, this cannot be guaranteed to succeed due the same reasons mentioned above.

In this mode, clients keep receiving data from the component, but cannot send any control commands to it. This mode allows users to receive a feedback that their action to move the component to mode `Off` has worked or not.

**`Inaccessible` mode**

As soon as a component moves to this mode, we assume that the component is not available on the network, and we do not attempt to read from it.

In this mode, clients do not receive any data from the component, and cannot send any control commands to it.

</details>

#### Proposal 4

**`Enabled` mode**

As soon as a component moves to this mode, we try to start it up. (Note that this cannot be guaranteed to succeed, since a component could refuse to start up, e.g., due to the component's internal configuration.)

In this mode, clients are allowed to read data from the component, and send commands to it. Sending commands to a component in this mode can be prevented by setting its respective bounds to (0,0), which would make the component read-only.

**`Disabled` mode**

As soon as a component moves to this mode, we try to stop it. (Again, this cannot be guaranteed to succeed due the same reasons mentioned above.)

In this mode,
- we will try to disconnect the component from the microgrid, by opening its relays.
- we will prevent applications from sending control commands to the component.
- we will keep getting data from this component.

### Use cases

- switch a component on - e.g., closing its AC/DC relays
- switch a component off - e.g., opening its AC/DC relays. Note that depending upon the component model, it may still be able to stay online and allow us to read from it.
- ~run maintenance activities on components  - e.g., battery calibration, firmware updates.~
- add a placeholder component to the component graph, to indicate its existence in case we cannot reach it over the network due to hardware-installation-related challenges.

_No response_

### Additional context

_No response_

tiyash-basu-frequenz:
This enum is used to explicitly define whether an electrical component is intended to be active (enabled) or inactive (disabled). This intention directly determines the component's availability to downstream services and applications.

Note that this enum differs from the `ElectricalComponentStatus` enum, which is used to report the actual operational state of the component.

Thanks to @thomas-nicolai-frequenz for providing the original version.

### DIFF
--- a/proto/frequenz/api/common/v1/microgrid/electrical_components/electrical_components.proto
+++ b/proto/frequenz/api/common/v1/microgrid/electrical_components/electrical_components.proto
@@ -182,6 +182,54 @@ enum ElectricalComponentStateCode {
   ELECTRICAL_COMPONENT_STATE_CODE_PRECHARGER_CLOSED = 42;
 }
 
+// `ElectricalComponentControlMode` explicitly defines whether an electrical
+// component is intended to be active (enabled) or inactive (disabled). This
+// intentional setting directly determines component availability to downstream
+// agents, applications, and services.
+//
+// !!! note "Control Mode & Component Availability"
+//     The Control Mode explicitly represents the user's or automation's
+//     intention to make the component either ACTIVE or INACTIVE. This directly
+//     impacts availability:
+//     - ACTIVE indicates the component is intended to be available and usable
+//       by downstream agents and services.
+//     - INACTIVE indicates the component is intended to be unavailable for use
+//       by downstream agents and services.
+//
+// !!! note "Relationship with Operational State"
+//     Control Mode explicitly represents intention, distinct from the
+//     component's actual internal Operational State.
+//     Operational State is independently reported by the component hardware
+//     (e.g., READY, CHARGING, DISCHARGING, ERROR).
+//
+// !!! caution "Immediate Operational Impact"
+//     Changing the Control Mode of an electrical component (e.g., via the
+//     Microgrid API) takes immediate effect on its availability to downstream
+//     agents and services. Setting a component to `INACTIVE` can thus instantly
+//     interrupt ongoing operations or cause service disruptions. Setting a
+//     component to `ACTIVE` may immediately connect it with other components,
+//     potentially causing electrical damage. Therefore, ensure changes are
+//     coordinated or planned appropriately to avoid unintended impact and/or
+//     damages.
+//
+// !!! example "Control Mode Change Impact"
+//     If a battery component's Control Mode is switched from `ACTIVE` to
+//     `INACTIVE`, it immediately becomes unavailable to downstream
+//     services and agents, potentially disrupting scheduled
+//     charging/discharging cycles.
+enum ElectricalComponentControlMode {
+  // The control mode is unspecified (should never be explicitly used).
+  ELECTRICAL_COMPONENT_CONTROL_MODE_UNSPECIFIED = 0;
+
+  // The component is explicitly enabled and intended to be available for
+  // downstream usage.
+  ELECTRICAL_COMPONENT_CONTROL_MODE_ACTIVE = 1;
+
+  // The component is explicitly disabled and intended to be unavailable for
+  // downstream usage.
+  ELECTRICAL_COMPONENT_CONTROL_MODE_INACTIVE = 2;
+}
+
 // A representation of all possible diagnostic codes that can occur for
 // electrical component, across all their categories. These diagnostic codes
 // can be used to refer to warnings or errors that are reported by the
--- a/proto/frequenz/api/common/v1/microgrid/electrical_components/electrical_components.proto
+++ b/proto/frequenz/api/common/v1/microgrid/electrical_components/electrical_components.proto
@@ -87,24 +87,6 @@ enum ElectricalComponentCategory {
   ELECTRICAL_COMPONENT_CATEGORY_HVAC = 14;
 }
 
-// ElectricalComponentStatus defines the possible statuses for a component.
-//
-// !!! note
-//     The status indicates the status set by the user via the user interface.
-//     The status is not yet included in the Component messages and should be
-//     added.
-//
-enum ElectricalComponentStatus {
-  // The status is unspecified. This should not be used.
-  ELECTRICAL_COMPONENT_STATUS_UNSPECIFIED = 0;
-
-  // The component is active.
-  ELECTRICAL_COMPONENT_STATUS_ACTIVE = 1;
-
-  // The component is inactive.
-  ELECTRICAL_COMPONENT_STATUS_INACTIVE = 2;
-}
-
 // Enum to represent the various states that a component can be in.
 // This enum is unified across all component categories for consistency.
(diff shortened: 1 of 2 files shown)


proto/frequenz/api/common/v1/microgrid/electrical_components/electrical_components.proto
```diff
@@ -121,6 +121,54 @@ enum ElectricalComponentStatus {
...
+// component is intended to be active (enabled) or inactive (disabled). This
+// intentional setting directly determines component availability to downstream
+// services and applications.
+//
+// !!! note "Control Mode & Component Availability"
+//     The Control Mode explicitly represents the user's or automation's
+//     intention to make the component either ACTIVE or INACTIVE. This directly
+//     impacts availability:
+//     - ACTIVE indicates the component is intended to be available and usable
+//       by downstream agents and services.
```
thomas-nicolai-frequenz, line 177: Maybe we say "agents, applications and services"?

proto/frequenz/api/common/v1/microgrid/electrical_components/electrical_components.proto
```diff
@@ -121,6 +121,54 @@ enum ElectricalComponentStatus {
...
+//     - INACTIVE indicates the component is intended to be unavailable and
+//       ignored by downstream agents and services.
```
thomas-nicolai-frequenz: I would say not usable instead of ignored.

proto/frequenz/api/common/v1/microgrid/electrical_components/electrical_components.proto
```diff
@@ -121,6 +121,54 @@ enum ElectricalComponentStatus {
...
+//     Microgrid API) takes immediate effect on its availability to downstream
+//     agents and services. Setting a component to `INACTIVE` can thus instantly
+//     interrupt ongoing operations or cause service disruptions. Setting a
+//     component to `ACTIVE` may immediately connect it with other components,
+//     potentially causing electrical damage. Therefore, ensure changes are
+//     coordinated or planned appropriately to avoid unintended impact and/or
+//     damages.
+//
+// !!! example "Control Mode Change Impact"
+//     If a battery component's Control Mode is switched from ACTIVE to
```
thomas-nicolai-frequenz: Is that only true for batteries?

proto/frequenz/api/common/v1/microgrid/electrical_components/electrical_components.proto
```diff
@@ -121,6 +121,54 @@ enum ElectricalComponentStatus {
...
+//     INACTIVE, it immediately becomes unavailable to downstream
+//     energy-management services, potentially disrupting scheduled
```
thomas-nicolai-frequenz: Lets remove the "energy-management" part and just say "services and agents".

proto/frequenz/api/common/v1/microgrid/electrical_components/electrical_components.proto
tiyash-basu-frequenz: Done

proto/frequenz/api/common/v1/microgrid/electrical_components/electrical_components.proto
tiyash-basu-frequenz: Done

proto/frequenz/api/common/v1/microgrid/electrical_components/electrical_components.proto
tiyash-basu-frequenz: I changed it to "unavailable for use", since the component should still be available, but no agent will be able to control it or use it.
Does the phrasing look okay now?

proto/frequenz/api/common/v1/microgrid/electrical_components/electrical_components.proto
tiyash-basu-frequenz: It should be true for every component that we can disable. As an example, battery works well though.

thomas-nicolai-frequenz (dismissed):

proto/frequenz/api/common/v1/microgrid/electrical_components/electrical_components.proto
thomas-nicolai-frequenz: ah its an example I missed that.

proto/frequenz/api/common/v1/microgrid/electrical_components/electrical_components.proto
thomas-nicolai-frequenz: yes.

llucax (approved):

PR #308, opened 2025-04-15
Add an `ElectricalComponentError` message

tiyash-basu-frequenz:
This commit adds the `ElectricalComponentError` message to the `electrical_components.proto` file. This message is designed to represent error or warning conditions reported by microgrid electrical components. It extends a standardized error code with contextual information useful for diagnostics or vendor-specific insights.

### DIFF
--- a/proto/frequenz/api/common/v1/microgrid/electrical_components/electrical_components.proto
+++ b/proto/frequenz/api/common/v1/microgrid/electrical_components/electrical_components.proto
@@ -461,37 +461,68 @@ enum ElectricalComponentErrorCode {
   ELECTRICAL_COMPONENT_ERROR_CODE_RELAY_CYCLE_LIMIT_REACHED = 60;
 }
 
+// Represents an error or warning condition reported by a microgrid electrical
+// component.
+//
+// This message extends a standardized error code with contextual information
+// useful for diagnostics, such as subsystem names, measured values, or
+// vendor-specific insights.
+//
+// !!! example "Typical Component Error"
+//     ```json
+//     {
+//       "errorCode": "ELECTRICAL_COMPONENT_ERROR_CODE_OVERTEMPERATURE",
+//       "vendorErrorCode": 2003,
+//       "message": "Temperature sensor 3 reported 61.5°C (limit: 60°C)"
+//     }
+//     ```
+//
+// !!! note "Component-Originated Diagnostics"
+//     This message is intended to reflect diagnostics originating from
+//     electrical components (e.g., inverters, batteries), either from internal
+//     firmware or via gateway logic.
+//
+// !!! note "Vendor-Specific Codes"
+//     When available, `vendor_error_code` will reflect manufacturer-defined
+//     error identifiers, typically exposed via Modbus or similar interfaces.
+//     These codes are vendor- and firmware-specific and must be interpreted
+//     using the vendor's official documentation.
+message ElectricalComponentError {
+  // A standardized error code representing the category of the issue.
+  ElectricalComponentErrorCode error_code = 1;
+
+  // Optional vendor-provided error code for the error, for vendor-specific
+  // insights or more granular diagnostics.
+  optional uint32 vendor_error_code = 2;
+
+  // Optional human-readable message providing additional context.
+  string message = 3;
+}
+
 // Representation of a component state and errors.
 message ElectricalComponentState {
   // The time at which the state was sampled.
   google.protobuf.Timestamp sampled_at = 1;
 
-  // List of states of the microgrid component.
+
+  // List of operational states currently active for the component.
   //
   // !!! note
-  //    The list will contain unique members. No state will exist twice in
-  //    this list.
+  //     The list must not contain duplicate state codes.
   repeated ElectricalComponentStateCode states = 2;
 
-  // List of warnings for the microgrid component.
+  // List of non-critical warnings detected for the component.
   //
   // !!! note
-  //    This list may have warnings even if the component state is not in an
-  //    error state.
-  //
-  // !!! note
-  //    The list will contain unique members. No warning will exist twice in
-  //    this list.
-  repeated ElectricalComponentErrorCode warnings = 3;
+  //     Warnings may be reported even when the component is operational.
+  //     No duplicate warning codes allowed.
+  repeated ElectricalComponentError warnings = 3;
 
-  // List of errors for the microgrid component.
+  // List of critical errors currently affecting the component.
   //
   // !!! note
-  //    This list is expected to have errors if and only if the component is in
-  //    an error state.
-  //
-  // !!! note
-  //    The list will contain unique members. No error will exist twice in
-  //    this list.
-  repeated ElectricalComponentErrorCode errors = 4;
+  //     This list is expected to be populated only when the component is in an
+  //     error state.
+  //     No duplicate error codes allowed.
+  repeated ElectricalComponentError errors = 4;
 }
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -13,6 +13,7 @@
 - Renamed `components` to `electrical_components` and related messages, fields, enums.
 - Added message linking microgrid and sensor IDs.
 - Added new message definitions for communication components.
+- Added new message `ElectricalComponentError` to represent errors in microgrid electrical components.
 
 ## Bug Fixes
 


thomas-nicolai-frequenz (dismissed):
LGTM!

proto/frequenz/api/common/v1/microgrid/electrical_components/electrical_components.proto
```diff
@@ -461,37 +461,62 @@ enum ElectricalComponentErrorCode {
...
+//       "vendorErrorCode": 2003,
+//       "message": "Temperature sensor 3 reported 61.5°C (limit: 60°C)"
+//     }
+//     ```
+//
+// !!! note "Component-Originated Diagnostics"
+//     This message is intended to reflect diagnostics originating from
+//     electrical components (e.g., inverters, batteries), either from internal
+//     firmware or via gateway logic.
+message ElectricalComponentError {
```
ktickner, line 490: @tiyash-basu-frequenz are we accounting for warnings at all here?

proto/frequenz/api/common/v1/microgrid/electrical_components/electrical_components.proto
```diff
@@ -461,37 +461,62 @@ enum ElectricalComponentErrorCode {
...
+  // A standardized error code representing the category of the issue.
+  ElectricalComponentErrorCode error_code = 1;
+
+  // Optional vendor-provided error code for the error, for vendor-specific
+  // insights or more granular diagnostics.
+  optional uint32 vendor_error_code = 2;
```
ktickner, line 496: I thought this was custom, which gave me the impression it wasn't exclusively vendor codes

proto/frequenz/api/common/v1/microgrid/electrical_components/electrical_components.proto
tiyash-basu-frequenz: We are using the same message for warnings as well. The message name is a bit misleading though.

proto/frequenz/api/common/v1/microgrid/electrical_components/electrical_components.proto
tiyash-basu-frequenz: IIUC we do not have a use-case for this beyond expressing vendor-specific error codes. So IMO it is okay if this is set specifically to this naming.

proto/frequenz/api/common/v1/microgrid/electrical_components/electrical_components.proto
```diff
@@ -461,37 +461,62 @@ enum ElectricalComponentErrorCode {
...
-  //    error state.
+  // List of non-critical warnings detected for the component.
   //
   // !!! note
-  //    The list will contain unique members. No warning will exist twice in
-  //    this list.
-  repeated ElectricalComponentErrorCode warnings = 3;
+  //     Warnings may be reported even when the component is operational.
+  //     No duplicate warning codes allowed.
+  repeated ElectricalComponentError warnings = 3;
```
ktickner, line 519: okay so warnings are still specified as the Error message... which kinda makes sense logically

> This list is expected to be populated only when the component is in an
  
  could warnings be reported alongside errors? Not relevant so much here, but how do you distinguish a warning from an error? based on the code?

proto/frequenz/api/common/v1/microgrid/electrical_components/electrical_components.proto
ktickner: yeah i saw that further down, sorry 😅

proto/frequenz/api/common/v1/microgrid/electrical_components/electrical_components.proto
ktickner: simple to change in the event we get a use case?
just Tom's wording was specifically custom, which flags to me that maybe it will be appropriate in the future.

thomas-nicolai-frequenz: Yeah, custom implies that it accepts a custom code that is not standardised across the Frequenz platform. Its usually vendor specific.

proto/frequenz/api/common/v1/microgrid/electrical_components/electrical_components.proto
thomas-nicolai-frequenz: yes, it can. Warnings are seperate from errors as both are stored in their respective variables called `warnings` and `errors`.

llucax:
Probably for another PR, but maybe it can be renamed to `ElectricalComponentIssue` to be less confusing when used as a warning?

proto/frequenz/api/common/v1/microgrid/electrical_components/electrical_components.proto
ktickner: > Its usually vendor specific.

exactly what im raising though - the property name would suggest it is always vendor specific

tiyash-basu-frequenz: I feel the challenge with the term `custom` is that it is not specific enough. If we have a number there, then the  user would not know just by looking at the data what it could mean. That means we would need to support it with more documentation, or perhaps more fields, or perhaps specifically worded error messages, which is not what we are doing right now.

On the other hand, the only use case we have now is supporting vendor-specific error codes. Naming it as such would remove the guesswork for the user.

That's why I would lean towards marking it as vendor specific, and try not to optimise for unknown future use cases. If the need comes, we can anyway extend the definition.

Just my 2 cents. Happy to hear other arguments :)

ktickner: in this case would the message still relate to the FQZ error code?
Wouldn't the user need to look up more information anyway? for example, what the vendor is? but i do get your point here

ktickner: i guess in a world where we want other custom codes, we may want to consider it being more robust, for example being able to specify where the code is coming from (meaning it would require work anyway, so might as well not think too hard right now)

thomas-nicolai-frequenz: We might want to add a note?
```
// ...
// !!! note "Vendor-Specific Codes"
//     If available, `vendor_error_code ` should reflect manufacturer-defined error identifiers,
//     typically exposed via Modbus or similar interfaces. These codes are implementation-specific
//     and must be interpreted using the vendor's official documentation or device register map.
message ElectricalComponentError {
// ...
```

tiyash-basu-frequenz:
> Probably for another PR, but maybe it can be renamed to `ElectricalComponentIssue` to be less confusing when used as a warning?

Sure. I have a few more contenders:
- `ElectricalComponentAnomaly`
- `ElectricalComponentDiagnostic`

Also, I would also extend this to the enum name as well.

Will create an issue depending upon how the discussions in this PR go.

ktickner:
I prefer issue or diagnostic.
`Anomaly` makes me think Cthulu

proto/frequenz/api/common/v1/microgrid/electrical_components/electrical_components.proto
tiyash-basu-frequenz: >Wouldn't the user need to look up more information anyway? for example, what the vendor is? but i do get your point here

IMO it is slightly different if you specify it as a vendor specific code or a custom one. If we mark it as a vendor-specific code, then a small static addendum, like Tom suggested, is enough. If it is custom, then you need to generate that context dynamically, which is more complex.

tiyash-basu-frequenz: @thomas-nicolai-frequenz would you prefer the extra documentation at a message level, or the field level? I guess there is a reason you specified it as a message level, but still just checking.

tiyash-basu-frequenz: Added it at the message level, with a few minor edits. Ready for a look.

thomas-nicolai-frequenz:
> ElectricalComponentDiagnostic

Would align well with other APIs.

thomas-nicolai-frequenz (approved):
LGTM

ktickner (approved):
LGTM

PR #310, opened 2025-04-16
Update sensor error representation

tiyash-basu-frequenz:


proto/frequenz/api/common/v1/microgrid/sensors/sensors.proto
```diff
@@ -161,6 +193,17 @@ message SensorState {
...
+  //
+  // !!! note
+  //    This list is expected to have warnings if and only if the sensor is
+  //    reporting warnings.
+  //
+  // !!! note
+  //    The list will contain unique members. No warning will exist twice in
+  //    this list.
+  repeated SensorDiagnostic warnings = 3;
+
```
ktickner, line 206: so one sensor could have multiple states, which contains multiple codes, with multiple errors and warnings?

tiyash-basu-frequenz: > so one sensor could have multiple states, which contains multiple codes

TL;DR - yes :)

We have seen that a few electrical components report states from their sub-componetns. E.g., batteries may provide a BMS state and a DC relay state, and EV chargers may provide an EV charger state and a cable state. What we did there is report the states in an array, so that we can flexibly and transparently report whatever states the electrical component is giving us.

We followed the same design approach for sensors. Hence the states are repeated.

> with multiple errors and warnings?

We have also seen that components can report multiple errors and warnings at the same time. Hence these need to be arrays too.

thomas-nicolai-frequenz (approved):
LGTM

PR #312, opened 2025-04-28
CI: Check py folders for init files

Marenz:


florian-wagner-frequenz:
Wouldn't it be better to habe this as a proper bash script which we could also add as a commit hook? This seems like it might lead to some thrashing in the review phase for something trivially checkable.

Marenz:
sure, I am not set on any specific solution, I just tried to find the lowest-hanging fruit that solves the problem (of not knowing why the documentation gen failed, due to a missing init file)

Marenz:
I mean, I am not expecting it to be a common occurrence to add definition all the time, thus I didn't invest too much time :)

florian-wagner-frequenz:
Fair point

llucax (commented):
I think generating the missing files is a better solution. I think it is probably less effort to implement than adding the ci (although of course you already added the ci, so if you want to merge it go ahead).

Also deployment to all repos is simpler, as no modifications are needed, just a bump of repo-config.

PR #331, opened 2025-05-15
Reorganize the contents of a few proto files

tiyash-basu-frequenz:
This commit reorganizes the contents of the following proto files:
- `communication_components.proto`
- `electrical_components.proto`
- `sensors.proto`

The reorganization includes:
- Moving the enums to the top of the file for better visibility and organization.
- Sequencing the messages in an arbitrary logical order, ensuring that related messages are grouped together.

This PR does not change the protobuf messages themselves in any way.


llucax (approved):
I didn't look to the diff as it is very hard to read, so I assume no code was changed, only things were moved around. I looked at the resulting files and LGTM.

PR #338, opened 2025-05-19
Rename `metric_sample.proto` to `metrics.proto`

tiyash-basu-frequenz:
This change renames the `metric_sample.proto` file to `metrics.proto` to better reflect its content. The file contains multiple metric-related messages, including `SimpleMetricValue` and `MetricSample`, which are used to represent metric values and samples, respectively. The new name `metrics.proto` is more descriptive and aligns with the purpose of the file.


thomas-nicolai-frequenz (approved):
LGTM!

PR #346, opened 2025-05-22
Remove `SensorMetricSample` in favour of `MetricSample`

### LINKED ISSUE #339: Remove `SensorMetricSample` in favour of `MetricSample`
### What's needed?

The [`SensorMetricSample`](https://github.com/frequenz-floss/frequenz-api-common/blob/c9a075d6b8a13d391651d29db9faaba527a3b6b2/proto/frequenz/api/common/v1/microgrid/sensors/sensors.proto#L210-L220) message is a subset of the [`MetricSample`](https://github.com/frequenz-floss/frequenz-api-common/blob/c9a075d6b8a13d391651d29db9faaba527a3b6b2/proto/frequenz/api/common/v1/metrics/metric_sample.proto#L187-L240) message. We do not see a scenario where these two messages will ever diverge, given how generic the `MetricSample` message is.

It does not add much value in having `SensorMetricSample`, since it is just a more specific version of a generic message type without offering unique fields or functionality that would necessitate a separate definition. Having `SensorMetricSample` in this scenario introduces unnecessary redundancy and complexity to the API's data model without providing a clear benefit or addressing a distinct use case beyond what `MetricSample` already handles.

### Proposed solution

One solution would be to remove `SensorMetricSample` and use `MetricSample` instead.

### Use cases

_No response_

_No response_

### Additional context

See #260 , #287, #324

tiyash-basu-frequenz:
This commit removes the `SensorMetricSample` message in favour of using `MetricSample` for sensors. The `SensorMetricSample` message was a subset of the `MetricSample` message.

From a design perspective focused on consistency, flexibility, and minimizing API surface area while handling variations via optional fields (which is idiomatic in Protobuf3), there's a strong argument that `SensorMetricSample` is redundant and potentially confusing, and that having only `MetricSample` is a cleaner design.

closes #339

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -35,6 +35,7 @@
 - Renamed `SensorState` to `SensorStateSnapshot` to better reflect its purpose.
 - Renamed `sampled_at` timestamps for state snapshots to `origin_time`.
 - Renamed `sampled_at` timestamps for metric samples to `sample_time`.
+- Remove `SensorMetricSample` in favour of using `MetricSample` for sensors.
 
 ## Bug Fixes
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -36,6 +36,7 @@
 - Renamed `sampled_at` timestamps for state snapshots to `origin_time`.
 - Renamed `sampled_at` timestamps for metric samples to `sample_time`.
 - Remove `SensorMetricSample` in favour of using `MetricSample` for sensors.
+- Remove `SensorMetric` enum, since it was unused and redundant.
 
 ## Bug Fixes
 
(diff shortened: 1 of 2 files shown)


llucax (commented):
Shouldn't this remove `SensorMetric `too?

tiyash-basu-frequenz:
> Shouldn't this remove `SensorMetric `too?

Done.

llucax (approved):
I also wonder if it is worth clarifying that for sensors bounds will always be empty in the `SensorData` message, but approving as it is something very minor.

tiyash-basu-frequenz:
>I also wonder if it is worth clarifying that for sensors bounds will always be empty in the SensorData message, but approving as it is something very minor.

I will remember to add it in the next PR.

PR #352, opened 2025-05-23
Remove `SensorCategory` enum

### LINKED ISSUE #348: `SensorCategory` is being used inconsistently w.r.t. practical use cases
### What's needed?

The `SensorCategory` enum has variants that do not apply to practical cases.

https://github.com/frequenz-floss/frequenz-api-common/blob/ebf585f580a277410a3619a6e7d5c2361d7960f4/proto/frequenz/api/common/v1/microgrid/sensors/sensors.proto#L18-L43

In practice, sensors contain different types of sensors (e.g., temperature and humidity), so they would belong to different sensor categories at the same time. However, they can belong to only one category at a time: https://github.com/frequenz-floss/frequenz-api-common/blob/ebf585f580a277410a3619a6e7d5c2361d7960f4/proto/frequenz/api/common/v1/microgrid/sensors/sensors.proto#L86.

This is inconsistent.

### Proposal 1

Allow a sensor to belong to multiple categories at the same time.

This would men changing the following: https://github.com/frequenz-floss/frequenz-api-common/blob/ebf585f580a277410a3619a6e7d5c2361d7960f4/proto/frequenz/api/common/v1/microgrid/sensors/sensors.proto#L86 to
```proto
    repeated SensorCategory categories = 4;
```

The issue here is that this is less of a category, and more of a list of supported metrics. Additionally, this can change dynamically, based on the sensor's states/errors and firmware version. So this will need to be renamed. (which would then be similar to implementing proposal 3, and adding it back a non-breaking change).

The other question with this proposal is if it even is a valid requirement to have this list or not. If a sensor is reporting metrics, then the list of available metrics can be obtained even from the data stream. If a sensor is not streaming any data (for whichever reason), then does having this list help in any way or not.

### Proposal 2

Replace all the variants in the `SensorCategory` enum with just one variant for now - `SENSOR_CATEGORY_ENVIRONMENTAL_SENSOR`, and ensure that all sensors that report only environmental/ambient data belong to this category.

After a bit of brainstorming, this is what I got n terms of a higher-level categorization:
```
Weather Station
Air Quality Monitoring Station (AQMS)
Water Quality Monitoring Station / Sonde
Indoor Environmental Quality (IEQ) Monitor
Soil Monitoring Station
```

Nearly all of these categories have use-cases in microgrids. So for this option, we replace all the existing variants in the `SensorCategory` enum with `WEATHER_STATION`. We limit it to just one variant for now, leaving the scope to extend it easily. We can remove it in the next major release if needed.

### Proposal 3

Remove the `SensorCategory` enum altogether, since we do not have a well-defined categorization at the moment. This allows the API to evolve without being constrained by a premature or inconsistent categorization, and a more suitable mechanism can be introduced later when the requirements for sensor categorization are clearer.

---

I am personally leaning towards proposal 3 for now.

### Use cases

_No response_

_No response_

### Additional context

_No response_

tiyash-basu-frequenz:
This enumeration was not useful and potentially confusing. Sensors can report different sensor metrics, and they could belong to several of these categories simultaneously. This defeats the purpose of having singular categories for sensors. We need to rethink how to categorize sensors. Until then, having it does not add any value, and therefore it has been removed.

closes #348

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -45,6 +45,7 @@
     - The code `UNDERVOLTAGE_SHUTDOWN` has been removed in favour of `UNDERVOLTAGE`.
     - New diagnostic codes have been added to cover more cases, especially for inverters.
     - The codes have been renumbered.
+- Removed `SensorCategory` enum, since it was not useful and potentially confusing. Sensors can report different sensor metrics, and they could belong to several of these categories simultaneously. This defeats the purpose of having singular categories for sensors. We need to rethink how to categorize sensors. Until then, having it does not add any value, and therefore it has been removed.
 
 ## Bug Fixes
 
--- a/proto/frequenz/api/common/v1/microgrid/sensors/sensors.proto
+++ b/proto/frequenz/api/common/v1/microgrid/sensors/sensors.proto
@@ -15,33 +15,6 @@ import "frequenz/api/common/v1/microgrid/lifetime.proto";
 
 import "google/protobuf/timestamp.proto";
 
-// Enumerated sensor categories.
-enum SensorCategory {
-  // Unspecified
-  SENSOR_CATEGORY_UNSPECIFIED = 0;
-
-  // Thermometer (temperature sensor)
-  SENSOR_CATEGORY_THERMOMETER = 1;
-
-  // Hygrometer (humidity sensor)
-  SENSOR_CATEGORY_HYGROMETER = 2;
-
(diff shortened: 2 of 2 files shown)


llucax (dismissed):

tiyash-basu-frequenz:
Rebased

llucax (approved):
Bye bye! 👋

thomas-nicolai-frequenz (commented):
LGTM

PR #355, opened 2025-06-02
Avoid warnings about protobuf gencode version

llucax:
We know about this warning and it is not a problem for us, as protobuf gurantees compatibility between two consecutive major versions.

When we really need to bump the dependency, we'll get an error that is not ignored.


llucax:
Will force-merge this, as the CI is broken without it.

PR #356, opened 2025-06-02
Bump protobuf and grpcio versions

llucax:


llucax:
Will force merge these, we are already taking it for other repos.

PR #372, opened 2025-06-13
Change type of `metrics.MetricSample.connection` field

### LINKED ISSUE #258: Extend definition for metric sources to include a category
### What's needed?

Here is the current definition of metric sources: https://github.com/frequenz-floss/frequenz-api-common/blob/ca56060fbf55975299233aa6af15b6dfa6e77bc6/proto/frequenz/api/common/v1/metrics/metric_sample.proto#L225-L239

The current implementation lacks the ability to identify the specific source of power, such as whether the source is linked to a battery, photovoltaic (PV) array, or another component. This omission creates ambiguity when multiple DC sources are connected.

For example, in a hybrid inverter with two DC strings—one from a battery and the other from a PV array—both strings provide DC power, but there is no clear way to distinguish between them based on the data. This differentiation is essential for monitoring and analyzing the behavior of each DC source independently, which could be crucial for operations.

### Proposal 1

Extend the definition to a message with a source category (the docs need to be properly written in the PR):

```proto
message MetricValueSourceCategory {
    // The metric value originates from its default source.
    // This could vary depending upon the component category and model.
    // E.g., by default, the DC_POWER metric for a battery inverter with one battery will originate from the single battery attached to the inverter.
    UNSPECIFIED = 0;

    // The metric value originates from a battery.
    BATTERY = 1;

    // The metric value originates from a PV array.
    PV = 2;
}

message MetricValueSource {
    MetricSourceCategory category = 0;
    string name = 1;
}

message MetricSample {
    ...
    MetricValueSource source = 5;
}
```

### Use cases

_No response_

### Workaround 1

Define a convention for source strings, e.g., metrics originating from batteries would always be named `dc_battery_n`, where `n` is a whole number.

The issue with this approach is that the API does not encode this specification, and it therefore depends fully on the client and server implementation. This opens up the risk of inconsistency and errors, especially if the convention is not strictly adhered to across different implementations, leading to potential confusion or misinterpretation of data.

### Additional context

_No response_

tiyash-basu-frequenz:
This commit changes the type of the `metrics.MetricSample.connection` field from `string` to a new message type `metrics.MetricConnection`.

This change allows for better categorization of the connection type, which can be useful in control and monitoring applications. The new `MetricConnection` message includes a `MetricConnectionCategory` enum that categorizes the connection type (e.g., battery, PV array, temperature`s), and a `name` field that can be used to identify the specific connection point from which the metric was obtained.

closes #258

### DIFF
--- a/proto/frequenz/api/common/v1/metrics/metrics.proto
+++ b/proto/frequenz/api/common/v1/metrics/metrics.proto
@@ -172,6 +172,50 @@ enum Metric {
   METRIC_SENSOR_IRRADIANCE = 166;
 }
 
+// Enumerated categories of connections from which metrics can be obtained for a
+// given electrical component or sensor.
+enum MetricConnectionCategory {
+  // Default value.
+  // Do not use this value.
+  METRIC_CONNECTION_CATEGORY_UNSPECIFIED = 0;
+
+  // A generic connection for metrics that do not fit into any other category.
+  METRIC_CONNECTION_CATEGORY_OTHER = 1;
+
+  // A connection to a metric representing a battery.
+  METRIC_CONNECTION_CATEGORY_BATTERY = 2;
+
+  // A connection to a metric representing a photovoltaic (PV) array or string.
+  METRIC_CONNECTION_CATEGORY_PHOTOVOLTAIC = 3;
+
+  // A connection to a metric representing ambient conditions.
+  METRIC_CONNECTION_CATEGORY_AMBIENT = 10;
+
+  // A connection to a metric representing a cabinet or an enclosure.
+  METRIC_CONNECTION_CATEGORY_CABINET = 11;
+
+  // A connection to a metric representing a heatsink.
+  METRIC_CONNECTION_CATEGORY_HEATSINK = 12;
+
+  // A connection to a metric representing a transformer.
+  METRIC_CONNECTION_CATEGORY_TRANSFORMER = 13;
+}
+
+// A connection to a metric representing from which a metric was obtained.
+message MetricConnection {
+  // The category of the connection from which the metric was obtained.
+  MetricConnectionCategory category = 1;
+
+  // A  string that can be used to identify the specific connection from which
+  // the metric was obtained.
+  //
+  // This is expected to be populated when the same `Metric` variant can be
+  // obtained from multiple distinct inputs or connection points on the
+  // component. Knowing the connection for the metric can help in certain
+  // control and monitoring applications.
+  string name = 2;
+}
+
 // Representation of a sampled metric along with its value.
 //
 // !!! note
@@ -222,7 +266,7 @@ message MetricSample {
   // ==== values here are allowed and will be accepted
   repeated Bounds bounds = 4;
 
-  // An optional string that can be used to identify the specific connection
+  // An optional field that can be used to identify the specific connection
   // from which the metric was obtained.
   //
   // This is expected to be populated when the same `Metric` variant can be
@@ -239,10 +283,20 @@ message MetricSample {
   //   to determine the SoC of the battery using its voltage, identifying the
   //   `dc_battery_0` connection is important.
   // - A sensor unit might report temperature from different connected probes
-  //   (e.g., `temp_sensor_outdoor`, `temp_sensor_indoor`).
+  //   (e.g., `temp_sensor_outdoor`, `temp_sensor_transformer`).
   //
-  // In cases where the component has just one such connection for a metric,
-  // or the context is otherwise unambiguous, this field is not expected to
-  // be present, because the connection is implicit.
-  optional string connection = 5;
+  // !!! note "Default Connection Behavior"
+  //     If this field is unset, it is implicitly assumed that the component
+  //     only has a single connection for this metric, and the connection is
+  //     uniquely identified by the component itself.
+  //
+  // !!! example
+  //     ```
+  //     // DC power from battery string 0 in a hybrid inverter
+  //     source {
+  //       category: METRIC_CONNECTION_CATEGORY_BATTERY
+  //       name: "dc_battery_0"
+  //     }
+  //     ```
+  optional MetricConnection connection = 5;
 }
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -73,6 +73,8 @@
 
 - The file `location.proto` has been moved to the `types` package, moving the message `Location` to `types.Location`.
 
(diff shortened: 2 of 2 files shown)


proto/frequenz/api/common/v1/metrics/metrics.proto
```diff
@@ -172,6 +172,49 @@ enum Metric {
...
+// given electrical component or sensor.
+enum MetricConnectionCategory {
+  // Default value.
+  METRIC_CONNECTION_CATEGORY_UNSPECIFIED = 0;
+
+  // Represents a connection to a battery.
+  METRIC_CONNECTION_CATEGORY_BATTERY = 1;
+
+  // Represents a connection to a PV array.
+  METRIC_CONNECTION_CATEGORY_PV_ARRAY = 2;
```
llucax: For inverter types we use `SOLAR`, do we really want to use something different here?

`SOLAR` always sounded weird to me though, not sure if it is a industry standard, or if we want to rename the inverter type too, and in that case if it makes sense to use `PV_ARRAY` or just `PV` or avoid acronyms and go for `PHOTOVOLTAIC`?

proto/frequenz/api/common/v1/metrics/metrics.proto
```diff
@@ -172,6 +172,49 @@ enum Metric {
...
+
+  // Represents a connection to a temperature sensor measuring the ambient
+  // temperature.
```
llucax: Do we want to be so specific? Can this be for something else than temperature? Like humidity, dew point, etc.? Same for CABINET, HEATSINK, etc.

proto/frequenz/api/common/v1/metrics/metrics.proto
```diff
@@ -172,6 +172,49 @@ enum Metric {
...
+  METRIC_CONNECTION_CATEGORY_AIR = 10;
```
llucax: Also if this means it comes from the "ambient", "AIR" sounds a bit too vague, wouldn't it make more sense to have something maybe like `AMBIENT`? `EXTERNAL_ENVIRONMENT`?

proto/frequenz/api/common/v1/metrics/metrics.proto
```diff
@@ -172,6 +172,49 @@ enum Metric {
...
+  // Represents a connection to a temperature sensor measuring the
+  //
+  // temperature of a transformer.
+  METRIC_CONNECTION_CATEGORY_TRANSFORMER = 13;
+}
+
+// Represents a connection from which a metric was obtained.
+message MetricConnection {
+  // The category of the connection from which the metric was obtained.
+  MetricConnectionCategory code = 1;
```
llucax: This also sounds inconsistent. I would either name `MetricConnectionCode` or name the field `category`.

proto/frequenz/api/common/v1/metrics/metrics.proto
```diff
@@ -239,10 +282,20 @@ message MetricSample {
...
   //
-  // In cases where the component has just one such connection for a metric,
-  // or the context is otherwise unambiguous, this field is not expected to
-  // be present, because the connection is implicit.
-  optional string connection = 5;
+  // !!! example
+  //     ```
+  //     // DC power from battery string 0 in a hybrid inverter
+  //     source {
+  //       category: BATTERY
```
llucax: Is this missing the whole prefix `METRIC_CONNECTION_CATEGORY_` on purpose? If it is done like this usually to keep examples short, LGTM.

llucax (commented):
So many questions...

proto/frequenz/api/common/v1/metrics/metrics.proto
tiyash-basu-frequenz: `PHOTOVOLTAIC` sounds best. Changed to that.

proto/frequenz/api/common/v1/metrics/metrics.proto
tiyash-basu-frequenz: I reworked the docs on this a bit. Maybe see how it looks now.

proto/frequenz/api/common/v1/metrics/metrics.proto
tiyash-basu-frequenz: I changed it to `AMBIENT`.

proto/frequenz/api/common/v1/metrics/metrics.proto
tiyash-basu-frequenz: Right, this should have been `category` all along.

proto/frequenz/api/common/v1/metrics/metrics.proto
tiyash-basu-frequenz: Added the prefix here.

proto/frequenz/api/common/v1/metrics/metrics.proto
llucax: Do you think it is worth changing in the inverter type too? (not in this PR)

proto/frequenz/api/common/v1/metrics/metrics.proto
```diff
@@ -172,6 +172,50 @@ enum Metric {
...
+  METRIC_CONNECTION_CATEGORY_CABINET = 11;
+
+  // A connection to a metric representing a heatsink.
+  METRIC_CONNECTION_CATEGORY_HEATSINK = 12;
+
+  // A connection to a metric representing a transformer.
+  METRIC_CONNECTION_CATEGORY_TRANSFORMER = 13;
+
+  // A generic connection for metrics that do not fit into any other category.
+  METRIC_CONNECTION_CATEGORY_OTHER = 100;
```
llucax: I missed this before, wouldn't it make more sense to put it as 1, next to `UNSPECIFIED` to avoid the gaps in the tags?

llucax (commented):
Only one more thing.

proto/frequenz/api/common/v1/metrics/metrics.proto
tiyash-basu-frequenz: good point, makes sense

tiyash-basu-frequenz: Done

llucax (approved):

proto/frequenz/api/common/v1/metrics/metrics.proto
tiyash-basu-frequenz: We should. I will create a separate PR for that.

tiyash-basu-frequenz: https://github.com/frequenz-floss/frequenz-api-common/pull/373

PR #375, opened 2025-06-16
Override the default include path when generating protobuf files

llucax:
By default `repo-config` will include the `submodules/frequenz-api-common` directory, which is us, and we don't want to include ourselves as a submodule, so we need to explicitly set the `include_paths` to an empty array.

We also need to bump the dependency on `frequenz-repo-config` to get a bug fix that allows overriding with an empty array.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@
 requires = [
   "setuptools == 80.9.0",
   "setuptools_scm[toml] == 8.3.1",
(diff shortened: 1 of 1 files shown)


llucax:
Draft because it depends on the `repo-config` v0.13.5 release.

llucax:
Ready for review.

tiyash-basu-frequenz (approved):

PR #380, opened 2025-06-19
Move all electrical component types to a single file

### LINKED ISSUE #377: Move `InverterType` and other `*Type` messages to `electrical_components.proto`
### What's needed?

Currently, messages like `InverterType`, `BatteryType`, `EvChargerType`, etc. are in their own files, and these files contain only these enums. This leads to a bit of back and forth between the files while reading the specs, which be optimised.

### Proposed solution

Move these enums to `electricakl_components.proto`

### Use cases

_No response_

_No response_

### Additional context

_No response_

tiyash-basu-frequenz:
This commit moves all the electrical component types from their individual files into `electrical_components.proto`.

closes #377


shsms (dismissed):

pytests/test_common_v1alpha7.py
```diff
@@ -136,71 +124,13 @@ def test_module_import_microgrid_electrical_components() -> None:
...
-    )
-
-    assert transformer_pb2_grpc is not None
+    # pylint: disable=no-member
+    assert electrical_components_pb2.Battery is not None
+    assert electrical_components_pb2.EvCharger is not None
+    assert electrical_components_pb2.Fuse is not None
+    assert electrical_components_pb2.GridConnectionPoint is not None
+    assert electrical_components_pb2.Inverter is not None
+    assert electrical_components_pb2.VoltageTransformer is not None
```
shsms: It looks like these tests are only for checking the generated code is reachable, we don't need to check individual structs are there.

tiyash-basu-frequenz: I removed these tests, keeping in line with the rest of the tests here.

shsms (approved):

PR #399, opened 2025-07-09
Rename voltage-transformer to power-transformer

### LINKED ISSUE #387: Review name and description of `VoltageTransformer`
https://github.com/frequenz-floss/frequenz-api-common/blob/6e0a5b14d7af66e89c41404d137908d2f1715e16/proto/frequenz/api/common/v1alpha7/microgrid/electrical_components/electrical_components.proto#L556-L561

This is more accurately described as a power transformer or simply a transformer.

A voltage transformer is a different thing used to lower voltage levels so that they can be measured: https://en.wikipedia.org/wiki/Voltage_transformer

I already mentioned the naming mismatch to @tiyash-basu-frequenz.

_Originally posted by @shsms in https://github.com/frequenz-floss/frequenz-client-microgrid-python/pull/165#discussion_r2182248086_

tiyash-basu-frequenz:
Voltage-transformers  are used in voltmeters for pretection, whereas what we needed was a power transformer, which is used for bulk transfer of electrical energy with voltage stepping up or down.

This renaming makes the electrical component name more consistent with the underlying concept of the component.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -18,6 +18,9 @@
   - `ElectricalComponentStateSnapshot.component_id` has been renamed to `ElectricalComponentStateSnapshot.electrical_component_id`
   - `microgrid.MicrogridComponentIds` has been renamed to `microgrid.MicrogridElectricalComponentIds`.
   - `MicrogridComponentIDs.component_ids` has been renamed to `MicrogridElectricalComponentIds.electrical_component_ids`.
+  - `electrical_components.VoltageTransformer` has been renamed to `electrical_components.PowerTransformer`, to make the name more consistent with the underlying concept.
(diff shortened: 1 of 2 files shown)


llucax (dismissed):
LGTM.

We are also using the term transformer in `metrics.proto` (`METRIC_INVERTER_TEMPERATURE_TRANSFORMER` and `METRIC_CONNECTION_CATEGORY_TRANSFORMER`). I guess this is s completely different concept, as it refers to the internals of a component, so it is OK that they don't match (and maybe even desirable to avoid confusion), but I thought I bring it up just in case.

tiyash-basu-frequenz:
I fixed the typo pointed out by copilot. Since the PR was otherwise approved, I'll merge it. :)

PR #400, opened 2025-07-09
Rename `type` package to `types` in v1alpha8 proto files

### LINKED ISSUE #392: Protobuf package name `type` causes awkward usage in Rust due to reserved keyword
### What's needed?

The protobuf package, originally named `types`, was [renamed to type](https://github.com/frequenz-floss/frequenz-api-common/issues/365) in version `v1alpha7` to align with Google's naming conventions, as seen in their `google/type` package.

While this change creates consistency with Google's APIs, it introduces a stylistic and ergonomic issue in Rust. Since `type` is a reserved keyword in Rust, developers must use the raw identifier `r#type` to reference the generated module.

For example, what would ideally be use `crate::frequenz::api::common::v1alpha7::type;` becomes `use crate::frequenz::api::common::v1alpha7::r#type;`.

This required use of `r#type` is not idiomatic in Rust and can be considered stylistically unclean. We should evaluate whether the benefit of aligning with Google's naming convention outweighs the negative impact on the developer experience in Rust.

### Proposed solution

Revert it back to types.

### Use cases

_No response_

_No response_

### Additional context

_No response_

tiyash-basu-frequenz:
This commit renames the `type` package to `types` in the v1alpha8 proto files to avoid using reserved keywords in programming languages.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -21,6 +21,7 @@
   - `electrical_components.VoltageTransformer` has been renamed to `electrical_components.PowerTransformer`, to make the name more consistent with the underlying concept.
   - `ElectricalComponentCategorySpecificInfo.kind.voltage_transformer` has been renamed to `ElectricalComponentCategorySpecificInfo.kind.power_transformer`, to make the name more consistent with the underlying concept.
   - `ElectricalComponentCategory.ELECTRICAL_COMPONENT_CATEGORY_VOLTAGE_TRANSFORMER` has been renamed to `ElectricalComponentCategory.ELECTRICAL_COMPONENT_CATEGORY_POWER_TRANSFORMER`, to make the name more consistent with the underlying concept.
+  - The `type` package has been renamed to `types`, to avoid using reserved keywords in programming languages.
 
 
 ## Bug Fixes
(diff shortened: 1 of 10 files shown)


llucax (approved):
I'm still not 100% convinced about this, as it is very complicated to make sure all symbols work in all languages, or put another way, the protoc for the language should deal with these problems by renaming keywords to something that can be used in that language (or using other language *hacks* like these raw identifiers in rust). And if you were to use the Google APIs, you'll still have this problem.

That said, given that rust is one of the primary languages we target, I'm OK with taking the practical approach and proceed with the rename, so approving.

llucax:
Removing from the queue in case you want to think about it :wink:

tiyash-basu-frequenz:
>I'm still not 100% convinced about this, as it is very complicated to make sure all symbols work in all languages, or put another way, the protoc for the language should deal with these problems by renaming keywords to something that can be used in that language (or using other language hacks like these raw identifiers in rust). And if you were to use the Google APIs, you'll still have this problem.


I appreciate the goal of aligning with Google's naming conventions, but I have reservations about changing `types` to `type`. My main concern is that it creates a new problem to solve a non-existent one.

Given this, I'm inclined to stick with `types`. It avoids introducing a known issue and seems like the more practical choice for our current workflow, which heavily focuses on rust.

PR #414, opened 2025-09-08
Add a new variant `Meter` to `ElectricalComponentCategorySpecificInfo`

tiyash-basu-frequenz:
This commit introduces a new message `Meter` to the `v1alpha8` package, which encapsulates static information about meters in a microgrid. Additionally, the `ElectricalComponentCategorySpecificInfo` enum has been extended to include the `Meter` variant. This enhancement enables the representation of meter-specific information within a microgrid.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,34 +2,16 @@
 
 ## Summary
 
-This release introduces the new `v1alpha8` version of the API, which includes several breaking changes compared to `v1alpha7`. The changes focus on improving consistency and clarity by renaming several symbols and removing unused components.
+<!-- Here goes a general summary of what this release is about -->
 
 ## Upgrading
 
-- A new package `frequenz.api.common.v1alpha8` has been introduced, containing the following breaking changes from `v1alpha7`.
+<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
 
-- Removed:
+## New Features
 
-    + `electrical_components.Fuse`
-    + `InverterType.INVERTER_TYPE_WIND_TURBINE`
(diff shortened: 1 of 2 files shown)


llucax (approved):
LGTM, only comment is it looks a bit like a *negative* saying "polarity reversed". I wonder if it would make more sense/be more extensible, if we have maybe a enum instead, like `SignConvention` with `PASSIVE` and `ACTIVE`.

When bouncing with Gemini, it agreed the enum is more robust but took it further and suggested `PASSIVE`, `REVERSED_PASSIVE` and `ACTIVE`, saying even if reversed passive and active are the same *intent matter*. I'm not so sure, but mentioning in case you can think of cases where differentiating both can make a difference.

In any case, this is just food for thought, I know in practice probably the enum works as well. But I thought throwing it out there in case anyone can think of more reasons to really prefer an enum in the long term.

FYI @shsms

llucax:
When challenging Gemini about the need for both it said (among other things):

> The difference in intent becomes critical for everything around that simple calculation: operations, maintenance, and risk management. Treating a mistake the same as a correct configuration introduces hidden risks that can cause serious problems down the line.
> 
> Think of the REVERSED_PASSIVE state as a "Check Engine" light for your meter. The car still drives, but the light tells you there's an underlying issue that needs to be tracked and eventually fixed.

So I guess for the UI/people on the other side of the SDK it could be valuable information, if there is really a conceptual difference between both.

tiyash-basu-frequenz:
@llucax thanks for the review. We are thinking about other ways to do this, which do not involve other ways to add this information to this API.

I will update the PR according to what we eventually agree upon.

tiyash-basu-frequenz:
We have successfully agreed upon an alternate solution. This PR s not needed anymore.

PR #469, opened 2026-04-23
Add model field to ElectricalComponent and Sensor messages

tiyash-basu-frequenz:
This PR adds a new field `model` to the `ElectricalComponent` and `Sensor` messages in the proto files. The `model` field includes both the manufacturer and the model name of the component or sensor.

Typically, the existing fields `manufacturer` and `model_name` are not used in separate contexts, despite being present in the structure for years. The new `model` field provides a more concise way to represent the model information, which is often used in practice.

In addition to the above, this PR deprecates the existing fields `manufacturer` and `model_name`, and these fields be removed in the next breaking change release.

### DIFF
--- a/proto/frequenz/api/common/v1alpha8/microgrid/sensors/sensors.proto
+++ b/proto/frequenz/api/common/v1alpha8/microgrid/sensors/sensors.proto
@@ -63,6 +63,10 @@ message Sensor {
 
   // The operational lifetime of the sensor.
   frequenz.api.common.v1alpha8.microgrid.Lifetime operational_lifetime = 7;
+
+  // The model name of the electrical component.
+  // Includes both the manufacturer and the model name.
+  string model = 8;
 }
 
 // Represents an error or warning condition reported by a microgrid sensor.
--- a/proto/frequenz/api/common/v1alpha8/microgrid/sensors/sensors.proto
+++ b/proto/frequenz/api/common/v1alpha8/microgrid/sensors/sensors.proto
@@ -55,11 +55,19 @@ message Sensor {
   // An optional name for the sensor.
   string name = 3;
 
-  // The sensor manufacturer.
-  string manufacturer = 5;
+  // The component manufacturer.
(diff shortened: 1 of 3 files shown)


llucax (commented):
LGTM, but I would keep the field IDs sorted, otherwise I think things can get very error prone if one needs to manually grep the whole message to see what's the latest/highest ID when a new field needs to be added.

llucax:
Hehe, that got officially confirmed, the current CI failure is actually because you got the ID wrong:

> proto/frequenz/api/common/v1alpha8/microgrid/electrical_components/electrical_components.proto:666:57: Field number 10 has already been used in "frequenz.api.common.v1alpha8.microgrid.electrical_components.ElectricalComponent" by field "model".

tiyash-basu-frequenz:
>LGTM, but I would keep the field IDs sorted, otherwise I think things can get very error prone if one needs to manually grep the whole message to see what's the latest/highest ID when a new field needs to be added.

True, not having them sorted is what caused the issue.

proto/frequenz/api/common/v1alpha8/microgrid/sensors/sensors.proto
```diff
@@ -55,11 +55,23 @@ message Sensor {
...
+  // The model name of the component.
+  //
+  // Note that this field is deprecated and should not be used.
+  // Please use the `model` field instead, which includes both the manufacturer
+  // and the model name.
+  string model_name = 6 [deprecated = true];
+
+  // The model name of the electrical component.
+  // Includes both the manufacturer and the model name.
+  string model = 8;
```
llucax, line 77: Don't you think it would be good to sort this one too? (so put it after `operational_lifetime = 7;`)

tiyash-basu-frequenz: done. Also rebased.

llucax (approved):

PR #470, opened 2026-04-23
Add operational mode to ElectricalComponent

tiyash-basu-frequenz:
This commit introduces a new field `operational_mode` to the `ElectricalComponent` message in the `electrical_components.proto` file.

This field can be used by clients to determine whether a component is active and operational, and whether it can be used for telemetry, control, or both.

### DIFF
--- a/proto/frequenz/api/common/v1alpha8/microgrid/electrical_components/electrical_components.proto
+++ b/proto/frequenz/api/common/v1alpha8/microgrid/electrical_components/electrical_components.proto
@@ -595,6 +595,30 @@ message ElectricalComponentCategorySpecificInfo {
   }
 }
 
+// This enum represents the operational mode of an electrical component,
+// indicating whether the component is active and operational, and whether it
+// provides telemetry data, accepts control commands, or both.
+enum ElectricalComponentOperationalMode {
+  // Default value when the operational mode is not explicitly set.
+  ELECTRICAL_COMPONENT_OPERATIONAL_MODE_UNSPECIFIED = 0;
+
+  // The component is inactive and not operational. It does not provide
+  // telemetry data, and it does not accept control commands.
+  ELECTRICAL_COMPONENT_OPERATIONAL_MODE_INACTIVE = 1;
+
+  // The component is active and operational. It only provides telemetry data,
+  // and it does not accept control commands.
+  ELECTRICAL_COMPONENT_OPERATIONAL_MODE_TELEMTRY_ONLY = 2;
+
+  // The component is active and operational. It only accepts control commands,
+  // and it does not provide telemetry data.
+  ELECTRICAL_COMPONENT_OPERATIONAL_MODE_CONTROL_ONLY = 3;
+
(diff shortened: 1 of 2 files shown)


proto/frequenz/api/common/v1alpha8/microgrid/electrical_components/electrical_components.proto
```diff
@@ -595,6 +595,30 @@ message ElectricalComponentCategorySpecificInfo {
...
+  // and it does not accept control commands.
+  ELECTRICAL_COMPONENT_OPERATIONAL_MODE_TELEMTRY_ONLY = 2;
+
+  // The component is active and operational. It only accepts control commands,
+  // and it does not provide telemetry data.
+  ELECTRICAL_COMPONENT_OPERATIONAL_MODE_CONTROL_ONLY = 3;
+
+  // The component is active and operational. It provides telemetry data and
+  // accepts control commands.
+  ELECTRICAL_COMPONENT_OPERATIONAL_MODE_CONTROL_AND_TELEMETRY = 4;
```
llucax, line 619: I guess this comes from something we are already using somewhere else, but I wonder if it doesn't make more sense to with a *flags* approach, as I guess code will want to check if one or the other is enabled to do something (either get data or send commands):

```proto
enum ElectricalComponentOperationalMode {
  // The component is active and operational. It only provides telemetry data,
  // and it does not accept control commands.
  ELECTRICAL_COMPONENT_OPERATIONAL_MODE_TELEMTRY = 0x1;

  // The component is active and operational. It only accepts control commands,
  // and it does not provide telemetry data.
  ELECTRICAL_COMPONENT_OPERATIONAL_MODE_CONTROL = 0x2;
}
```

And then we could do `operational_mode` optional if we want to keep an explicit *unspecified* option, and maybe rename it to `operational_mode_flags` if we want to be more explicit that these are flags.

Without this, we'll have to do a lot of `if ELECTRICAL_COMPONENT_OPERATIONAL_MODE_TELEMTRY_ONLY or ELECTRICAL_COMPONENT_OPERATIONAL_MODE_CONTROL_AND_TELEMETRY: get_telemetry()`, which seems overkill.

Not a biggie though, since I could also adapt this in the client, but if this will be the most common pattern everywhere maybe we want to just encode it in the protocol instead.

llucax (approved):
Approving in case you want to merge as is, but unless we I'm missing some good reason for the current approach, I think the flags make more sense.

proto/frequenz/api/common/v1alpha8/microgrid/electrical_components/electrical_components.proto
tiyash-basu-frequenz: Thanks for the feedback.

The pattern in the PR was decided upon a few rounds of discussion in other areas, and was found to be most suitable for exposing active/inactive and control/telemetry information to support general requirements that might go beyond what we support in the frequenz-floss org.

So I would prefer to keep it as is. If needed, we can change it in the future.

PR #471, opened 2026-04-24
dependabot: Add a migration script for grpc updates

llucax:
Dependabot fails to update grpc/protobuf dependencies correctly, as it limits itself to update build dependencies, but leave the runtime dependencies untouched.

This doesn't work because protobuf/grpc need to be generated with a version that is equals or older to the runtime version, so if we only bump the build dependency, a project might still depend on an older version at runtime, and we should not allow that.

This PR adds an auto-migration workflow to fix this, while also ensuring the major version for protobuf is bumped accordingly when a major version bump occurs.

This is an experimental change. If this work the upgrade script will be moved to a separate repository and the workflow added to repo-config.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -29,6 +29,11 @@ updates:
         exclude-patterns:
           # pydoclint has shipped breaking changes in patch updates often
           - "pydoclint"
+          # These need a migration script to fix dependabot missing updating
+          # the runtime dependencies
+          - "grpcio"
+          - "grpcio-tools"
+          - "protobuf"
       minor:
         update-types:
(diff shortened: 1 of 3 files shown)


llucax:
Marking as draft because I will try to do some tests in my fork first.

llucax:
Ready for review. Tested in:

* https://github.com/llucax/frequenz-api-common/pull/25
* https://github.com/llucax/frequenz-api-common/pull/26
* https://github.com/llucax/frequenz-api-common/pull/27

tiyash-basu-frequenz (approved):
Nice, thanks!

PR #497, opened 2026-06-10
Add common grid flexibility definitions

thomas-nicolai-frequenz:
## Summary

This PR adds common grid definitions for flexibility-related concepts shared across APIs.

It introduces common Protobuf definitions for:

* flexibility service types,
* provision durations and provision periods,
* typed domain resource URIs.

These types are intended to be reused by APIs such as SignalStream, FlexMarket, GridLink, and GridGuard, instead of duplicating API-specific definitions.

## Notes

The new definitions are placed under the common `grid` package because they describe grid-facing flexibility service participation and operator/resource identification concepts, while still being reusable by market-facing APIs where needed.

### DIFF
--- a/proto/frequenz/api/common/v1alpha8/grid/flexibility_service.proto
+++ b/proto/frequenz/api/common/v1alpha8/grid/flexibility_service.proto
@@ -1,0 +1,171 @@
+// Frequenz definitions for grid-facing flexibility services.
+//
+// Copyright 2025 Frequenz Energy-as-a-Service GmbH
+//
+// Licensed under the MIT License (the "License");
+// you may not use this file except in compliance with the License.
+
+syntax = "proto3";
+
+package frequenz.api.common.v1alpha8.grid;
+
+// Defines the type of flexibility service associated with a flexibility
+// resource.
+//
+// A flexibility resource can represent an individual asset, an asset group, a
+// microgrid, a Gridpool, or another market-specific aggregation that provides
+// or consumes flexibility.
+//
+// Each service type represents a flexibility service, often corresponding to
+// a market product or grid-operator requirement, with specific rules for
+// activation, delivery, telemetry, and control-signal handling.
+//
+// !!! note "Naming conventions"
+//     This enum uses region-neutral terminology and maps it to common
+//     flexibility-service names used by grid system operators and market
+//     operators in different regions.
+//
+// !!! note "Service behavior"
+//     This enum classifies the service type only. Service-specific behavior
+//     such as setpoint cadence, ramping windows, activation timing, telemetry
+//     requirements, and dispatch method is defined by the service or API that
+//     consumes this enum.
+//
+enum FlexServiceType {
+  // UNSPECIFIED: Default unspecified value — must not be used.
+  FLEX_SERVICE_TYPE_UNSPECIFIED = 0;
+
+  // PRIMARY: Primary Regulation
+  //
+  // Also known as:
+  // - Frequency Containment Reserve (FCR) [Europe]
+  // - Primary Frequency Response [US]
+  // - Dynamic Containment [UK]
+  //
+  // Provides autonomous frequency response within an assigned ΔP capacity band.
+  // Participating resources adjust active power locally and in real time based
+  // on measured frequency deviations.
+  //
+  // No centralized real-time dispatch signal is required for activation. The
+  // assigned capacity band may be determined by a bid, contract, schedule,
+  // or other service commitment.
+  //
+  // During the provision period, the assigned capacity responsibility may be
+  // updated, for example when availability changes or assets are deregistered.
+  FLEX_SERVICE_TYPE_PRIMARY = 1;
+
+  // SECONDARY_POS: Secondary Upward Regulation
+  //
+  // Also known as:
+  // - Automatic Frequency Restoration Reserve Up (aFRR+) [Europe]
+  // - Regulation Up [US, PJM]
+  // - Dynamic Regulation Up [UK]
+  //
+  // Supports the grid by increasing net export or reducing net import:
+  // - increasing generation or battery discharge,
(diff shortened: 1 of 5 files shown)


llucax:
CI failure:

```
INFO    -  Doc file 'SUMMARY.md' contains an unrecognized relative link 'protobuf-reference/', it was left as is.
INFO    -  Doc file 'SUMMARY.md' contains an unrecognized relative link 'python-reference/', it was left as is.
ERROR   -  mkdocstrings: frequenz.api.common.v1alpha8.identity.resource_uri_pb2 could not be found
ERROR   -  Error reading page 'python-reference/frequenz/api/common/v1alpha8/identity/resource_uri_pb2.md':
ERROR   -  Could not collect 'frequenz.api.common.v1alpha8.identity.resource_uri_pb2'
```

You need to create the placeholder file `py/frequenz/api/common/v1alpha8/identity/__init__.py` with just the standard copyright and docstring, like:

```py
# License: MIT
# Copyright © 2025 Frequenz Energy-as-a-Service GmbH

"""Frequenz common identity API v1alpha8."""
```

proto/frequenz/api/common/v1alpha8/identity/resource_uri.proto
```diff
@@ -0,0 +1,159 @@
...
+//     }
+//     ```
+//
+message ResourceUri {
+  // Full URI string (e.g., "group://frequenz/primary/pool-1").
+  string uri = 1;
+
+  // Resource type category, such as access owner, operator, or flexibility
+  // group.
+  ResourceUriType type = 2;
```
tiyash-basu-frequenz, line 166: This `type` field is the schema, right? As in, it is the first part of the URI string.

If yes, then wouldn't putting an additional `type` firld be redundant, and may be a place of potential mistaken divergence in code?

If we still want to keep this foels, then I woul dsuggest to rename it to `ResourceUriSchema` to align it with existing URI naming schemes.

proto/frequenz/api/common/v1alpha8/grid/provision_duration.proto
```diff
@@ -0,0 +1,97 @@
...
+// markets, this corresponds to the delivery block or product block for which
+// a flexibility service is contracted, committed, or delivered.
+//
+// !!! note "Duration vs. resolution"
+//     This field defines the duration of a single provision period. In market
+//     terms, this is often the delivery block or product block for the
+//     flexibility service. It does not define the cadence or resolution of
+//     telemetry samples, control instructions, or setpoint updates.
+//
+enum ProvisionDuration {
```
tiyash-basu-frequenz, line 32: Can we maybe add to the enum description why we have an enum instead of a flexible integer field for specifying a duration?

proto/frequenz/api/common/v1alpha8/grid/flexibility_service.proto
```diff
@@ -0,0 +1,171 @@
...
+  // - increasing generation or battery discharge,
+  // - decreasing consumption or battery charging.
+  //
+  // Secondary upward regulation is typically activated through recurring
+  // explicit control instructions or ΔP setpoints.
+  //
+  // Unlike primary regulation, activation is signal-based rather than purely
+  // autonomous. Participating resources are expected to follow received control
+  // instructions within the required ramping and delivery constraints.
+  FLEX_SERVICE_TYPE_SECONDARY_POS = 2;
```
tiyash-basu-frequenz, line 74: Is there a reason to call the enum variant `SECONDARY_POS` instead of maybe `SECONDARY_UPWARD_RESERVE`, which would be more in line with the docs?

proto/frequenz/api/common/v1alpha8/grid/provision_duration.proto
thomas-nicolai-frequenz: Because they are defined by the markets. These are standards. One can't just define something arbitrary.

thomas-nicolai-frequenz: I'll add it to clarify.

proto/frequenz/api/common/v1alpha8/identity/resource_uri.proto
thomas-nicolai-frequenz: The field is intentionally used for validation rather than as a second source of identity. The URI string carries the concrete resource identity, while `type` declares the expected resource category. 

I’ll clarify the field comment so it is clear that `type` is not meant to be the URI scheme itself, but the expected domain-specific resource type used for validation.

proto/frequenz/api/common/v1alpha8/grid/flexibility_service.proto
thomas-nicolai-frequenz: I would like to kept `POS` / `NEG` intentionally because the enum values align with the API’s grid-perspective sign convention: positive means increasing net export or reducing net import, while negative means reducing net export or increasing net import.

Using `UPWARD_RESERVE` / `DOWNWARD_RESERVE` would be more descriptive in isolation, but it would make the enum names much longer and would need to be applied consistently across secondary and manual reserve values. Since the comments already spell out the upward/downward meaning, I’d prefer to keep the compact suffixes tbh.

proto/frequenz/api/common/v1alpha8/grid/provision_duration.proto
tiyash-basu-frequenz: This helps. Thanks a lot.

proto/frequenz/api/common/v1alpha8/identity/resource_uri.proto
tiyash-basu-frequenz: Thanks!

tiyash-basu-frequenz (approved):

